PR #12, opened 2020-03-11
Removing unneeded functions

alicevhenshaw:


rbcp18:
Removing unnecessary functions from contact, in regards to next upgrade

PR #14, opened 2020-03-12
Modifying depositCollateralBorrower to accept ETH

salazarguille:
It contains:

- Fixing compilation errors in master.
- Modifying depositCollateralBorrower to accept ETH.
- Adding unit tests for depositCollateralBorrower.
- Adding CircleCI configuration.

image

image

### DIFF
--- a/contracts/ZeroCollateral.sol
+++ b/contracts/ZeroCollateral.sol
@@ -78,8 +78,7 @@ contract ZeroCollateralMain is ERC20Detailed, ERC20, MinterRole, Chainlink {
         uint256 lastBorrowId;
         uint256 amountPaidRedemptionPool;
         uint256 amountPaidDefaultPool;
-        uint256 collateralRedeemable;
-        uint256 collateralNonRedeemable;
+        uint256 collateral;
     }
 
     // array of all borrower accounts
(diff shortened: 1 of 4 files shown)


contracts/ZeroCollateral.sol
```diff
@@ -343,7 +343,19 @@ contract ZeroCollateralMain is ERC20Detailed, ERC20, MinterRole, Chainlink {
         }
     }
 
+    function getCollateralNeeded(uint256 amountBorrow) view public returns(uint256){
+        uint256 poolContributions = (borrowerAccounts[msg.sender].amountPaidRedemptionPool);
+        uint256 amountBorrowDecimals = amountBorrow.mul(10**6); // convert to DAI decimals
```
alicevhenshaw: Maybe I've misunderstood, but I believe DAI has 18 decimals not 6. The contract has a constant called `DAI_DECIMALS` storing `10**18` for this purpose

contracts/ZeroCollateral.sol
```diff
@@ -343,7 +343,19 @@ contract ZeroCollateralMain is ERC20Detailed, ERC20, MinterRole, Chainlink {
         }
     }
 
+    function getCollateralNeeded(uint256 amountBorrow) view public returns(uint256){
+        uint256 poolContributions = (borrowerAccounts[msg.sender].amountPaidRedemptionPool);
```
alicevhenshaw: As discussed in a message, it would be better to have `borrowerAddress` as a parameter instead of using `msg.sender`

contracts/ZeroCollateral.sol
```diff
@@ -404,6 +416,58 @@ contract ZeroCollateralMain is ERC20Detailed, ERC20, MinterRole, Chainlink {
         return true;
     }
 
+    function calculateInterestDiscount(uint256 amountBorrow, uint256 numberDays) view public returns (uint256) {
+        // amountBorrow > 0 
+        require(amountBorrow > 0, "Amount Borrow: LESS THAN 0");
+        
+        // calculate % of interest paid (8 decimals) + collateral (8 decimals), divided by borrow amount (2 decimals). Resulting 6 decimals.
+        uint256 x = borrowerAccounts[msg.sender].amountPaidRedemptionPool.add(borrowerAccounts[msg.sender].collateral).div(amountBorrow);
```
alicevhenshaw: Again it would be better to have a parameter than use `msg.sender`

contracts/ZeroCollateral.sol
salazarguille: You're right. The DAI has 18 decimals. This function was added again because the contract didn't compile, but the full contract will be reviewed to clean not needed functions, and issues like this.

contracts/ZeroCollateral.sol
salazarguille: I totally agree. Also, it depends on what you allow users to execute. For example: if you allow a user to get your getCollateralNeeded data, for sure, you need to add a new address parameter (and not use msg.sender). If you don't want to allow that, you need to use the msg.sender. 

We should analyze it and if it is needed to make the changes in a new task.

PR #15, opened 2020-03-13
clean code

haythem96:
Simple PR that: 
- clean code
- organize functions
- add reverts messages


contracts/ZeroCollateral.sol
```diff
@@ -131,12 +131,64 @@ contract ZeroCollateralMain is ERC20Detailed, ERC20, MinterRole, Chainlink {
         decimals
     )
     {
+        require((daiAddress != address(0)) && (daoAddress != address(0)), "ZeroCollateral::constructor - invalid addresses");
+
         daiContract = daiAddress;
         zcDaoContract = daoAddress;
+
+        borrowCount = 0;
```
alicevhenshaw: All uint256 initialise as 0 on creation of the contract storage, so you only need the initialisation of `blockAccruedInterest`

contracts/ZeroCollateral.sol
```diff
@@ -131,12 +131,64 @@ contract ZeroCollateralMain is ERC20Detailed, ERC20, MinterRole, Chainlink {
         decimals
     )
     {
+        require((daiAddress != address(0)) && (daoAddress != address(0)), "ZeroCollateral::constructor - invalid addresses");
```
alicevhenshaw: This revert message is over 32 characters long and so has to be handled in a special way. This means that the gas costs are more expensive. It would be better to try and have a message that was 32 characters or less.

alicevhenshaw: Same with the other revert messages

salazarguille: Right. To avoid potential similar issues, we may define a naming convention / standard errors codes or similar.

haythem96:
closed as most of the modifications are done in other PRs.

PR #17, opened 2020-03-13
Separating token, oracle and main contract

alicevhenshaw:


salazarguille:
I think the unit tests are failing.

PR #26, opened 2020-03-21
Implementing LenderInfo contract

salazarguille:
It contains:

- Most of the functions for the LenderInfo contract.
- Unit tests for all implemented functions (code coverage ~=95%) in LenderInfo contract.
- Prettier Solidity formatter plugin.
- Husky hooks:
   - Pre-commit: Formatting smart contracts.
   - Pre-push: Running unit tests.

The interest earned calculation process/formula is being analyzed by the team. So, it eventually will be changed in a new PR once it is defined.

image

image

I have used the Slither tool to analyze the LenderInfo.sol contract. It got only one issue related to the Solidity version. So, there weren't low/medium/high priority issues.

image

### DIFF
--- a/contracts/base/LenderInfo.sol
+++ b/contracts/base/LenderInfo.sol
@@ -33,9 +33,6 @@ contract LenderInfo is LenderInfoInterface {
 
     /* State Variables */
 
-    // total accrued interest
-    uint256 public totalAccruedInterest;
-
     // last block number of accrued interest
     uint256 public blockAccruedInterest;
 
@@ -55,6 +52,11 @@ contract LenderInfo is LenderInfoInterface {
 
     modifier isDaiPool(address anAddress) {
         require(areAddressesEqual(address(daiPool), anAddress), "Address has no permissions.");
+        _;
+    }
+
+    modifier isValid(address anAddress) {
+        anAddress.requireNotEmpty("Address is required.");
         _;
     }
 
@@ -78,7 +80,8 @@ contract LenderInfo is LenderInfoInterface {
     function zDaiTransfer(address sender, address recipient, uint256)
         external
         isZDai(msg.sender)
-        // TODO Check sender/recipient is not 0x0.
+        isValid(sender)
+        isValid(recipient)
         returns (bool)
     {
         // Updating accrued interest for zDAI sender.
@@ -93,7 +96,7 @@ contract LenderInfo is LenderInfoInterface {
     function zDaiMinted(address recipient, uint256)
         external
         isDaiPool(msg.sender)
-        // TODO Check recipient is not 0x0.
+        isValid(recipient)
         returns (bool)
     {
         // Updating accrued interest for zDAI recipient.
@@ -105,7 +108,7 @@ contract LenderInfo is LenderInfoInterface {
     function zDaiBurnt(address recipient, uint256)
         external
         isDaiPool(msg.sender)
-        // TODO Check recipient is not 0x0.
+        isValid(recipient)
         returns (bool)
     {
         // Updating accrued interest for zDAI recipient.
@@ -117,11 +120,26 @@ contract LenderInfo is LenderInfoInterface {
     function withdrawInterest(address recipient, uint256 amount)
         external
         isDaiPool(msg.sender)
-        // TODO Check recipient is not 0x0.
+        isValid(recipient)
         returns (uint256)
     {
+        // Updating accrued interest for zDAI recipient.
+        updateAccruedInterestFor(recipient);
 
-        return 0;
+        uint256 amountToWithdraw = amount;
+        uint256 accruedInterest = lenderAccounts[recipient].totalAccruedInterest;
+        
+        if (accruedInterest >= amountToWithdraw) {
+            lenderAccounts[recipient].totalAccruedInterest = accruedInterest.sub(amountToWithdraw);
+        } else {
+            amountToWithdraw = lenderAccounts[recipient].totalAccruedInterest;
+            lenderAccounts[recipient].totalAccruedInterest = 0;
+        }
+        lenderAccounts[recipient].lastBlockAccrued = this.getCurrentBlockNumber();
+
+        emit AccruedInterestWithdrawn(recipient, amountToWithdraw);
+
+        return amountToWithdraw;
     }
 
     function getCurrentBlockNumber()
--- a/contracts/base/LenderInfo.sol
+++ b/contracts/base/LenderInfo.sol
@@ -142,7 +142,7 @@ contract LenderInfo is LenderInfoInterface {
         return amountToWithdraw;
     }
 
-    function getCurrentBlockNumber() external view returns (uint256 blockNumber) {
+    function getCurrentBlockNumber() external view returns (uint256) {
         return block.number;
     }
 
@@ -184,7 +184,7 @@ contract LenderInfo is LenderInfoInterface {
         uint256 previousBlockAccruedInterest,
         uint256 currentBlockNumber,
         uint256 currentZDaiBalance
-    ) internal pure returns (uint256 newAccruedInterest) {
+    ) internal pure returns (uint256) {
         uint256 blocksDifference = currentBlockNumber.sub(previousBlockAccruedInterest);
         return currentAccruedInterest.add(blocksDifference.mul(currentZDaiBalance));
     }
@@ -199,7 +199,7 @@ contract LenderInfo is LenderInfoInterface {
 
     /** Private Functions */
 
-    function getZDaiBalanceOf(address anAddress) private view returns (uint256 balance) {
+    function getZDaiBalanceOf(address anAddress) private view returns (uint256) {
         return zdai.balanceOf(anAddress);
     }
 }
--- a/contracts/base/LenderInfo.sol
+++ b/contracts/base/LenderInfo.sol
@@ -88,25 +88,16 @@ contract LenderInfo is LenderInfoInterface {
 
         // Updating accrued interest for zDAI recipient.
         updateAccruedInterestFor(recipient);
-
     }
 
-    function zDaiMinted(address recipient, uint256)
-        external
-        isDaiPool(msg.sender)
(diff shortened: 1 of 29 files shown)


contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,216 @@
...
+        blockAccruedInterest = block.number;
+    }
+
+    /** External Functions */
+
+    function zDaiTransfer(address sender, address recipient, uint256)
+        external
+        isZDai(msg.sender)
+        isValid(sender)
+        isValid(recipient)
```
alicevhenshaw: The ZDAI.transfer will already check that the sender and recipient addresses are non-zero. See here:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cdf655f770cf8412fbd4d6aef3c55194a1233ef1/contracts/token/ERC20/ERC20.sol#L155

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,216 @@
...
+        // Updating accrued interest for zDAI recipient.
+        updateAccruedInterestFor(recipient);
+
+        return true;
+    }
+
+    function zDaiMinted(address recipient, uint256)
+        external
+        isDaiPool(msg.sender)
+        isValid(recipient)
```
alicevhenshaw: Same here for ZDAI.mint

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,216 @@
...
+        // Updating accrued interest for zDAI recipient.
+        updateAccruedInterestFor(recipient);
+
+        return true;
+    }
+
+    function zDaiBurnt(address recipient, uint256)
+        external
+        isDaiPool(msg.sender)
+        isValid(recipient)
```
alicevhenshaw: And same here for burn

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,216 @@
...
+        require(zdaiAddress != address(0x0), "ZDai address is required.");
+        require(daiPoolAddress != address(0x0), "Dai pool address is required.");
+        zdai = ZDAIInterface(zdaiAddress);
+        daiPool = DAIPoolInterface(daiPoolAddress);
+        blockAccruedInterest = block.number;
+    }
+
+    /** External Functions */
+
+    function zDaiTransfer(address sender, address recipient, uint256)
```
alicevhenshaw: If the `uint256` isn't needed in the contract then feel free to remove it from both the contract and the interface too. I just added it here in case it was needed.

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,216 @@
...
+        return true;
+    }
+
+    function withdrawInterest(address recipient, uint256 amount)
+        external
+        isDaiPool(msg.sender)
+        isValid(recipient)
+        returns (uint256)
+    {
+        // Updating accrued interest for zDAI recipient.
```
alicevhenshaw: `zDAI recipient` -> `recipient`. They aren't receiving ZDAI I believe? I think they're receiving DAI?

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,216 @@
...
+        uint256 amountToWithdraw = amount;
+        uint256 accruedInterest = lenderAccounts[recipient].totalAccruedInterest;
+        
+        if (accruedInterest >= amountToWithdraw) {
+            lenderAccounts[recipient].totalAccruedInterest = accruedInterest.sub(amountToWithdraw);
+        } else {
+            amountToWithdraw = lenderAccounts[recipient].totalAccruedInterest;
+            lenderAccounts[recipient].totalAccruedInterest = 0;
+        }
+        lenderAccounts[recipient].lastBlockAccrued = this.getCurrentBlockNumber();
```
alicevhenshaw: For what reason have you created `this.getCurrentBlockNumber` as a function?
It feels to me that it would be a lot simpler to simply have
```
lenderAccounts[recipient].lastBlockAccrued = block.number;
```

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,216 @@
...
+        uint256 previousBlockAccruedInterest = lenderAccounts[lender].lastBlockAccrued;
+        uint256 currentBlockNumber = this.getCurrentBlockNumber();
+
+        // Update last block accrued.
+        lenderAccounts[lender].lastBlockAccrued = currentBlockNumber;
+
+        // Update total accrued interest.
+        lenderAccounts[lender].totalAccruedInterest = calculateNewAccruedInterestFor(
+                                                            lenderAccounts[lender].totalAccruedInterest,
+                                                            previousBlockAccruedInterest,
```
alicevhenshaw: Might just be github, but this looks like very strange indentation?

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,216 @@
...
+        internal
+        view
+        returns (bool)
+    {
+        return leftAddress.isEqualTo(rightAddress);
+    }
+
+    /** Private Functions */
+
+    function getZDaiBalanceOf(address anAddress) private view returns (uint256 balance) {
```
alicevhenshaw: Again this function to me seems like it doesnt add anything - as it's not doing any form of calculation etc. It's doing 1 simple call . What is your reason for putting it inside its own function?

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,216 @@
...
+
+    function calculateNewAccruedInterestFor(
+        uint256 currentAccruedInterest,
+        uint256 previousBlockAccruedInterest,
+        uint256 currentBlockNumber,
+        uint256 currentZDaiBalance
+    )
+        internal
+        pure
+        returns (uint256 newAccruedInterest)
```
alicevhenshaw: Naming the return parameter declares a variable called that. In this case you don't use the return parameter name, so you should remove the name. It's fine to name the return parameter if you are then using that variable within the function.

contracts/base/LenderInfo.sol
salazarguille: Okay, I will remove it.

salazarguille: Okay, I will remove it.

contracts/base/LenderInfo.sol
salazarguille: Okay, I will remove it. Just to be on sync, our spec: "the zDaiMinted function will be executed by the DaiPool contract."

contracts/base/LenderInfo.sol
salazarguille: Okay, I will remove it. Just to be on sync, we need to verify the contract/s will invoke this function (and zDaiMinted).

contracts/base/LenderInfo.sol
salazarguille: Perfect. Let me confirm it.

contracts/base/LenderInfo.sol
salazarguille: Yes, you're right

contracts/base/LenderInfo.sol
salazarguille: It makes the code cleaner/simpler, and it allows me to mock the function (so, mocking the block.number) for the unit tests (and create different scenarios). This is one way to test contracts when they are using timestamps/block numbers/etc. Also, you can use the dependency 'mock-contract' (from Gnosis), but it depends on what you are testing.

contracts/base/LenderInfo.sol
salazarguille: Yes, you are right. I will configure a linter to fix these issues.

contracts/base/LenderInfo.sol
salazarguille: As I mentioned before, mainly to make the code simpler/readable/testable.

contracts/base/LenderInfo.sol
alicevhenshaw: Any update on this?

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,206 @@
...
+        isZDai(msg.sender)
+        returns (bool)
+    {
+        // Updating accrued interest for zDAI sender.
+        updateAccruedInterestFor(sender);
+
+        // Updating accrued interest for zDAI recipient.
+        updateAccruedInterestFor(recipient);
+
+        return true;
```
alicevhenshaw: If these functions can never return false, what is the reason for them returning a bool?

contracts/base/LenderInfo.sol
```diff
@@ -0,0 +1,206 @@
...
+
+        if (accruedInterest >= amountToWithdraw) {
+            lenderAccounts[recipient].totalAccruedInterest = accruedInterest.sub(
+                amountToWithdraw
+            );
+        } else {
+            amountToWithdraw = lenderAccounts[recipient].totalAccruedInterest;
+            lenderAccounts[recipient].totalAccruedInterest = 0;
+        }
+        lenderAccounts[recipient].lastBlockAccrued = this.getCurrentBlockNumber();
```
alicevhenshaw: This lastBlockAccrued update is already done in the function `updateAccruedInterestFor` in this function call

contracts/base/LenderInfo.sol
alicevhenshaw: Any update on this?

salazarguille: Yes, as the team is analyzing how the interest will be calculated, I think the amount will be required. But the new calculation will be developed in a new PR once It is defined.
I have created a task so we can track it.

contracts/base/LenderInfo.sol
salazarguille: You're right. I will remove it and in other functions in this contract too.

contracts/base/LenderInfo.sol
salazarguille: Good catch! Thanks!

PR #29, opened 2020-03-30
Adding DaiPool.depositDai contract/function

salazarguille:
It contains
- DaiPool.depositDai function
- Unit tests for implementation.
- Changes for migrations: adding DaiPool and Loans contracts.
- Some configuration (per network) for deployment.
- Loans contract: based on the husky pre-commit hook Loans was formatted based on prettier solidity formatter.

image
image
image


config/networks/ganache/chainlink.js
```diff
@@ -0,0 +1,6 @@
+
+module.exports = {
+    dataContracts: {
+        eth_usd: '0x1234567890123456789012345678901234567890',
```
alicevhenshaw: What is the use of this fake address right now? Should we not be putting in the correct contracts?

contracts/base/ZDai.sol
```diff
@@ -19,23 +19,16 @@ pragma solidity 0.5.17;
 import "openzeppelin-solidity/contracts/token/ERC20/ERC20Mintable.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/ERC20Detailed.sol";
 
+
 contract ZDai is ERC20Detailed, ERC20Mintable {
+    string private constant NAME = "ZeroCollateral DAI";
```
alicevhenshaw: Do we want a space between `zero collateral`?

contracts/mock/base/InitializableModifiersMock.sol
```diff
@@ -0,0 +1,37 @@
...
+import "../../base/Initializable.sol";
+
+/**
+    This contract is created ONLY for testing purposes.
+ */
+contract InitializableModifiersMock is Initializable {
+    
+    /** State Variables */
+
+    /** Connstructor */
```
alicevhenshaw: typo

test/utils/contracts.js
```diff
@@ -8,4 +8,10 @@ module.exports = {
             lenderData.lastAccruedInterest
         );
     },
+    createInstance: async newPromise => {
+        const contract = await newPromise;
+        assert(contract);
```
alicevhenshaw: For what reason is `assert(contract)` and `assert(contract.address)` needed? If the contract fails to deploy, then `await new ContractName()` fails and throws an error. What do these lines catch?

contracts/base/Initializable.sol
```diff
@@ -0,0 +1,53 @@
...
+// Libraries
+
+// Commons
+
+// Interfaces
+
+contract Initializable {
+    /* State Variables */
+
+    bool public _isInitialized;
```
alicevhenshaw: I believe that it is common practice for a variable beginning with `_` tends to be private/internal, and that public variables do not start with _. Do you follow a different convention?

contracts/base/DAIPool.sol
```diff
@@ -0,0 +1,139 @@
...
+
+    LoansInterface public loanInfo;
+
+    /** Modifiers */
+
+    /* Constructor */
+
+    /** External Functions */
+
+    function initialize(
```
alicevhenshaw: `Initializable` and `DAIPool` both have different functions with the same name: `initialize`. I think it is best to name these different things. Line 65 is making a call to the same function name that it is within.

contracts/base/DAIPool.sol
```diff
@@ -0,0 +1,139 @@
...
+        address zdaiAddress,
+        address daiAddress,
+        address lenderInfoAddress,
+        address loanInfoAddress
+    ) external isNotInitialized() {
```
alicevhenshaw: The code is inconsistent whether it uses parentheses on modifiers. e.g. line 59 uses `isNotInitialized()` and 78 uses `isInitialized`

contracts/base/DAIPool.sol
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+        @notice It allows users to deposit DAIs into the pool.
+        @dev the user must call DAI.approve function previously.
+        @param amount of DAIs to deposit in the pool.
+    */
+    function depositDai(uint256 amount) external isInitialized {
+        // Require DAI approval from sender to this contract.
+        requireDAIAllowance(msg.sender, amount);
```
alicevhenshaw: This function reverts if the allowance is not big enough. However `dai.transferFrom` will already revert if the allowance is not big enough. I think that `requireDAIAllowance` is a superfluous function as `transferFrom` will already check this.

contracts/base/DAIPool.sol
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+        @notice It transfers an amount of DAI tokens from an address to this contract.
+        @param from address where the DAI tokens will transfer from.
+        @param amount to be transferred.
+        @dev It throws a require error if 'transferFrom' invocation fails.
+     */
+    function daiTransferFrom(address from, uint256 amount) private {
+        bool transferFromResult = dai.transferFrom(from, address(this), amount);
+        require(transferFromResult, "Transfer from was not successful.");
```
alicevhenshaw: dai.transferFrom already fails if the transfer is not successful - you do not need to check this :)

contracts/base/DAIPool.sol
```diff
@@ -0,0 +1,139 @@
...
+    /**
+        @notice It mints ZDAI tokens, and send them to a specific address.
+        @param to address which will receive the minted tokens.
+        @param amount to be minted.
+        @dev This contract must has a Minter Role in ZDAI (mintable) token.
+        @dev It throws a require error if mint invocation fails.
+     */
+    function zdaiMint(address to, uint256 amount) private {
+        bool mintResult = zdai.mint(to, amount);
+        require(mintResult, "Mint was not successful.");
```
alicevhenshaw: `zdai.mint` also already reverts if it fails, and otherwise succeeds, so you do not need to check it has succeeded

contracts/base/Initializable.sol
salazarguille: You're right. I made it private.

contracts/base/DAIPool.sol
salazarguille: You're right, I have added the parentheses on them.

config/networks/ganache/chainlink.js
salazarguille: Each network has a configuration (ZC contracts, tokens address, and others). In this case, the configuration is for Chainlink, but it is not available in Ganache (also for example the token DAI).

That configuration is used to run Truffle scripts that interact with already deployed contracts.

I will create a dummy address (const) in a consts.js file. So, it can be imported in the conf files (and update the conf value when the contract is deployed).

contracts/base/DAIPool.sol
salazarguille: Why it would be an issue? Contracts can have overloaded functions (https://solidity.readthedocs.io/en/v0.5.15/contracts.html#function-overloading) without problems.

contracts/base/DAIPool.sol
alicevhenshaw: We agreed that checking the bool on transferFrom and mint is good in case we add other tokens in future

contracts/base/DAIPool.sol
alicevhenshaw: We agreed that checking the bool on transferFrom and mint is good in case we add other tokens in future

contracts/base/DAIPool.sol
alicevhenshaw: I know for a while truffle had an issue compiling overloaded names, but it may well be that that problem is now gone 👍

contracts/base/DAIPool.sol
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+        @notice It transfers an amount of DAI tokens from an address to this contract.
+        @param from address where the DAI tokens will transfer from.
+        @param amount to be transferred.
+        @dev It throws a require error if 'transferFrom' invocation fails.
+     */
+    function daiTransferFrom(address from, uint256 amount) private {
+        bool transferFromResult = dai.transferFrom(from, address(this), amount);
+        require(transferFromResult, "Transfer from was not successful.");
```
alicevhenshaw, line 114: This revert message is 33 characters - revert messages are cheaper if you can keep then within 32 characters. Maybe remove the word 'was'?

contracts/mock/base/InitializableModifiersMock.sol
```diff
@@ -0,0 +1,37 @@
...
+/**
+    This contract is created ONLY for testing purposes.
+ */
+contract InitializableModifiersMock is Initializable {
+    
+    /** State Variables */
+
+    /** Constructor */
+
+    function _externalIsNotInitialized() isNotInitialized() external {}
```
alicevhenshaw, line 29: For what reason do these functions begin with an underscore?

salazarguille: When I test an internal function I used _ in the mock contract.

Now I have changed it, I use 'external' instead of '_'.

PR #35, opened 2020-04-16
Rename LenderInfo and DAIPool contracts

salazarguille:
DAIPool => LendingPool: we will use different tokens in our pools (such as USDC and DAI in the first version).
LenderInfo => Lenders: so it is on sync with Loans contract.

It includes:
- Contract names.
- Variables and parameters in contracts and JS files.
- JS test file names.
- Event names.
- New deployment addresses.

image


contracts/base/LendingPool.sol
```diff
@@ -23,27 +23,27 @@ import "../util/ZeroCollateralCommon.sol";
...
-contract DAIPool is DAIPoolInterface, Initializable {
+contract LendingPool is LendingPoolInterface, Initializable {
     using AddressLib for address;
 
     /* State Variables */
 
-    IERC20 public dai;
+    IERC20 public token;
 
     ZDAIInterface public zdai;
```
alicevhenshaw, line 44: If we are now calling it the `lendingPool` not the `daiPool` should `zDAI` be renamed too?

contracts/interfaces/LendingPoolInterface.sol
```diff
@@ -0,0 +1,79 @@
...
+    /**
+        @notice Once a loan is liquidated, it transfers the amount of tokens to the liquidator address.
+        @param amount of tokens to liquidate.
+        @param liquidator address to receive the tokens.
+        @dev It throws a require error if this contract hasn't enough token balance.
+     */
+    function liquidationPayment(uint256 amount, address liquidator) external;
+
+    /**
+        @notice Ønce the loan is created, it transfers the amount of tokens to the borrower.
```
alicevhenshaw, line 42: Typo: `Ønce the loan`

contracts/base/LendingPool.sol
salazarguille: Good catch!

contracts/base/LendingPool.sol
```diff
@@ -0,0 +1,217 @@
...
+        @notice It initializes the contract state variables.
+        @param zTokenAddress zToken token address.
+        @param tokenAddress ERC20 token address.
+        @param lendersAddress Lenders contract address.
+        @param loansAddress Loans contract address.
+        @dev It throws a require error if the contract is already initialized.
+     */
+    function initialize(
+        address zTokenAddress,
+        address tokenAddress,
```
alicevhenshaw, line 75: I think maybe to differentiate between `zToken` and `token` more clearly, they should be called `zToken` and `lendingToken` ?

PR #40, opened 2020-04-24
Take Out Loan & Loan Terms Consensus

salazarguille:
It includes:

- New library to manage values (numbers array).
- Changes in Loan.takeOut
- New Loans.requestLoan / Loans.setLoanTerms

Note: The flow has changed. This PR is just to re-use code. So, some unit tests were disabled. The unit tests and new flow will be added in the next PRs.

image

### DIFF
--- a/contracts/base/Loans.sol
+++ b/contracts/base/Loans.sol
@@ -20,12 +20,12 @@ pragma experimental ABIEncoderV2;
 import "../interfaces/LoansInterface.sol";
 import "../interfaces/PairAggregatorInterface.sol";
 import "../interfaces/LendingPoolInterface.sol";
+import "../interfaces/LoanTermsConsensusInterface.sol";
 import "../util/ZeroCollateralCommon.sol";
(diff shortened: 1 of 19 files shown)


contracts/base/Loans.sol
```diff
@@ -41,26 +41,50 @@ contract Loans is LoansInterface, SignerRole {
     // At any time, this variable stores the next available loan ID
     uint256 public loanIDCounter;
 
+    uint256 public timeWindowToTakeOutLoan = 7 days;
+
     PairAggregatorInterface public priceOracle;
+
     LendingPoolInterface public lendingPool;
 
+    LoanTermsConsensusInterface public loanTermsConsensus;
```
hensha256, line 50: I think that can just be stored as an address because the only place it is used is in a modifier. The actual interface is never needed

salazarguille:
The take out loan & loan terms flow changed. So it will be implemented in other PRs.

PR #41, opened 2020-04-24
Adding Numbers library

salazarguille:
It contains:

- New numbers library to calculate min/max/sum/length.
- Unit tests.

image

### DIFF
--- a/contracts/util/NumbersList.sol
+++ b/contracts/util/NumbersList.sol
@@ -6,7 +6,7 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 library NumbersList {
     using SafeMath for uint256;
 
-    uint256 private constant PERCENTAGE_TO_DECIMAL = 100;
+    uint256 private constant PERCENTAGE_TO_DECIMAL = 10000;
 
     struct Values {
         uint256 length;
--- a/contracts/util/NumbersList.sol
+++ b/contracts/util/NumbersList.sol
@@ -6,7 +6,7 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 library NumbersList {
     using SafeMath for uint256;
 
-    uint256 private constant PERCENTAGE_TO_DECIMAL = 10000;
+    uint256 private constant PERCENTAGE_TO_DECIMAL = 100;
 
     struct Values {
         uint256 length;
@@ -19,7 +19,7 @@ library NumbersList {
         if (self.max < newValue) {
             self.max = newValue;
         }
-        if (self.min > newValue) {
+        if (self.min > newValue || self.length == 0) {
             self.min = newValue;
         }
         self.sum = self.sum.add(newValue);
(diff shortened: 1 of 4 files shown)


contracts/util/NumbersList.sol
```diff
@@ -0,0 +1,64 @@
...
+    uint256 private constant PERCENTAGE_TO_DECIMAL = 10000;
+
+    struct Values {
+        uint256 length;
+        uint256 max;
+        uint256 min;
+        uint256 sum;
+    }
+
+    function addValue(Values storage self, uint256 newValue) internal returns (bool) {
```
hensha256, line 18: `returns (bool)`
This does not return a bool

contracts/util/NumbersList.sol
```diff
@@ -0,0 +1,64 @@
...
+    function isWithinTolerance(Values storage self, uint256 tolerance)
+        internal
+        view
+        returns (bool)
+    {
+        uint256 average = getAverage(self);
+        uint256 toleranceAmount = average.mul(tolerance).div(PERCENTAGE_TO_DECIMAL);
+
+        uint256 minTolerance = average.sub(toleranceAmount);
+        if (self.min <= minTolerance) {
```
hensha256, line 54: I think these should maybe be strict inequality. i.e. `self.min < minTolerance`. If the acceptable tolerance is 800 +/- 200 lets say, then 600 should probably be accepted -> it's 599 thats not acceptable.

contracts/util/NumbersList.sol
```diff
@@ -0,0 +1,64 @@
...
+        uint256 max;
+        uint256 min;
+        uint256 sum;
+    }
+
+    function addValue(Values storage self, uint256 newValue) internal returns (bool) {
+        if (self.max < newValue) {
+            self.max = newValue;
+        }
+        if (self.min > newValue) {
```
hensha256, line 22: I think this should be:
 `if (self.min > newValue || self.length == 0) {`

Currently if you have an empty struct, and you try to add the value 2, then you will end up with the following:
```
Values {
        length = 1
        max = 2
        min = 0
        sum = 2
    }
```
Because the initial value of 0 is less than 2

salazarguille: I have added new unit tests and fixed some issues.

contracts/util/NumbersList.sol
```diff
@@ -6,7 +6,7 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 library NumbersList {
     using SafeMath for uint256;
 
-    uint256 private constant PERCENTAGE_TO_DECIMAL = 10000;
+    uint256 private constant PERCENTAGE_TO_DECIMAL = 100;
```
hensha256, line 9: This should be `10000`. The percentage `tolerance` has 2 decimal places added to it. So if we want tolerance of 1.25%, `tolerance = 125` and the decimal we want is `0.0125` so you have to divide by `10000`

PR #42, opened 2020-04-28
Migration: Support Multiple Pools

salazarguille:
It contains:

- Migration supports multiple pools (DAI and USDC).
- Fix in Chainlink Oracle.
- Unit tests to check whether collateral sent is enough or not.
- Configuration improvement.
- Scripts: Refactoring lending pool scripts. Changes in loans (scripts) are pending (they are being changed).

image


contracts/base/Loans.sol
```diff
@@ -361,4 +361,15 @@ contract Loans is LoansInterface, SignerRole {
...
     }
+
+    function isCollateralSentEnough(
+        uint256 msgValue,
+        uint256 amountToBorrow,
+        uint256 collateralRatio
+    ) internal view returns (bool) {
+        uint256 oneLendingTokenPriceWeis = uint256(priceOracle.getLatestAnswer());
+        uint256 collateralSent = msgValue.div(oneLendingTokenPriceWeis);
+        uint256 collateralNeeded = amountToBorrow.mul(collateralRatio).div(1000);
```
hensha256, line 372: this should be `.div(10000);`
the contract has a constant `TEN_THOUSAND` for this purpose. This is due to the fact that collateral ratio is a percentage with 2 decimals. So a collateralRatio of 65.43% is stored as 6543. Therefore to find the collateral needed you must do `amountToBorrow * 0.6543`. Which is `amountToBorrow.mul(collateralRatio).div(10000);`

contracts/base/Loans.sol
```diff
@@ -361,4 +361,15 @@ contract Loans is LoansInterface, SignerRole {
             liquidated: false
         });
     }
+
+    function isCollateralSentEnough(
+        uint256 msgValue,
+        uint256 amountToBorrow,
+        uint256 collateralRatio
+    ) internal view returns (bool) {
+        uint256 oneLendingTokenPriceWeis = uint256(priceOracle.getLatestAnswer());
```
hensha256, line 370: I believe there is a disconnect between the values in this contract. Some are looking at USDC in whole tokens, and some are looking at USDC in units of tokens.

`oneLendingTokenPriceWeis` does not have the token decimals incorporated yet. From the screenshot you shared yesterday of the oracle response, at a price of 200 USDC for 1 ETH, the contract would return the following:
1 USDC = 5000000000000000 WEI = 0.005 ETH.
However this is 1 whole USDC, not 1 unit of USDC.
So we have `oneLendingTokenPriceWeis = 5000000000000000`
Then if 2 ETH was sent, `collateralSent = (2 ETH / 0.005 ETH) = 400` => i.e. 400 USDC worth

However, `amountToBorrow` is used in the Loans contract as follows:
`lendingPool.createLoan(amountBorrow, msg.sender);`
which in turn performs:
`tokenTransfer(borrower, amount);`
This means that `amountToBorrow` IS incorporating the units of the token. If you want to have 400 whole USDC, you have to call `lendingPool.createLoan(400*(10**6), msg.sender)`.

contracts/base/Loans.sol
```diff
@@ -361,4 +361,15 @@ contract Loans is LoansInterface, SignerRole {
             liquidated: false
         });
     }
+
+    function isCollateralSentEnough(
```
hensha256, line 365: This function isnt used anywhere? Was it meant to be incorporated to the actual contract?

contracts/base/Loans.sol
salazarguille: Okay, perfect. I was using 3 digits for the collateral ratio. I have changed it.

contracts/base/Loans.sol
salazarguille: Now, it considers the lending token decimals to verify whether collateral sent is enough or not.

contracts/base/Loans.sol
salazarguille: This function can be used in the takeOutLoan function (where the borrower sends the collateral).

As I changed the configuration for the Chainlink Oracle, I tested it before using it (in takeout loan).

contracts/base/Loans.sol
```diff
@@ -362,14 +364,25 @@ contract Loans is LoansInterface, SignerRole {
...
         uint256 msgValue,
         uint256 amountToBorrow,
         uint256 collateralRatio
     ) internal view returns (bool) {
+        // Calculate Collateral Sent (in lending tokens) = ether sent / 1 lending token price in ether -wei(s)- * a whole lending token (with decimals).
+        uint256 aWholeLendingToken = getAWholeLendingToken();
         uint256 oneLendingTokenPriceWeis = uint256(priceOracle.getLatestAnswer());
-        uint256 collateralSent = msgValue.div(oneLendingTokenPriceWeis);
-        uint256 collateralNeeded = amountToBorrow.mul(collateralRatio).div(1000);
+        uint256 collateralSent = msgValue.div(oneLendingTokenPriceWeis).mul(
```
hensha256, line 380: I would recommend doing the multiplication before doing the division in solidity, as division loses accuracy.
e.g. in solidity:
10.mul(6).div(3) = 20 but
10.div(3).mul(6) = 18

salazarguille: Good catch!

contracts/base/Loans.sol
```diff
@@ -361,4 +363,26 @@ contract Loans is LoansInterface, SignerRole {
             liquidated: false
         });
     }
+
+    function getAWholeLendingToken() internal view returns (uint256) {
```
hensha256, line 367: as internal functions should `getAWholeLendingToken` and `isCollateralSentEnough` begin with underscores?

migrations/utils/PoolDeployer.js
```diff
@@ -0,0 +1,54 @@
...
+        ChainlinkPairAggregator
+    } = this.artifacts;
+
+    await this.deployer.deployWith(`ChainlinkPairAggregator_${aggregatorName.toUpperCase()}`, ChainlinkPairAggregator, aggregatorAddress, txConfig);
+    await this.deployer.deployWith(`LendingPool_${zTokenName}`, LendingPool, txConfig);
+    await this.deployer.deployWith(`InterestConsensus_${zTokenName}`, InterestConsensus, requiredSubmissions, maximumTolerance, txConfig);
+    await this.deployer.deployWith(`Lenders_${zTokenName}`, Lenders, ZToken.address, LendingPool.address, InterestConsensus.address, txConfig);
+    await this.deployer.deployWith(`Loans_${zTokenName}`, Loans, ChainlinkPairAggregator.address, LendingPool.address, txConfig);
+  
+    const daiLendingPoolInstance = await LendingPool.deployed();
```
hensha256, line 38: Should the word `dai` be used here? Most of this file seems to be token-agnostic. SO maybe it should be `lendingPoolInstance`

PR #50, opened 2020-05-29
Feature/compound integration

hensha256:


salazarguille:
Please, add a description about the changes, and the unit tests passing.

contracts/interfaces/CErc20Interface.sol
```diff
@@ -0,0 +1,28 @@
+pragma solidity 0.5.17;
```
salazarguille, line 1: I think it should be in "contracts/providers/compound/CErc20Interface.sol".

contracts/base/LendingPool.sol
```diff
@@ -164,6 +179,9 @@ contract LendingPool is Base, LendingPoolInterface {
         isLoan()
         whenLendingPoolNotPaused(address(this))
     {
+        // Withdraw the tokens from compound
```
mark1323, line 182: Pre-existing problem, but it seems like we should be depositing DAI tokens on liquidation, not withdrawing them.  ETH should be withdrawn and sent to liquidator.

contracts/interfaces/CErc20Interface.sol
mark1323: I agree - seems Compound-specific

hensha256: The reason I put it in the interfaces folder is because 'PairAggregatorInterface' is in the interfaces folder too (which is chainlink). Happy to move it but was going for consistency

config/networks/ropsten/compound.js
```diff
@@ -0,0 +1,6 @@
+const { DUMMY_ADDRESS } = require('../../consts');
+
+module.exports = {
+    CDAI: DUMMY_ADDRESS,
```
salazarguille, line 4: Should we add the address in ropsten?

salazarguille:
Good job! I just left a comment about the ropsten conf.

contracts/base/LendingPool.sol
```diff
@@ -231,11 +271,6 @@ contract LendingPool is Base, LendingPoolInterface {
         require(transferFromResult, "TransferFrom wasn't successful.");
     }
 
-    function requireEnoughLendingTokenBalance(uint256 amount) private view {
```
mark1323, line 234: Why is this being taken out?  Is it because all the funds are all in Compound, and a withdrawal from Compound that's "too much" will simply fail?  We may want a better way of handling this in the future (e.g. failure with a reason that's communicated to the user), depending on what the final withdrawal mechanism will be.

hensha256: yes I could potentially add it back in. I try to prioritise reducing the number of unnecessary checks, as users don't like to use apps that have high gas costs. But if you'd like this back in because of the error message then I can do so

PR #51, opened 2020-05-29
Adding Integration tests and scripts

salazarguille:
It contains:

- New integration tests (1_liquidate_by_collateral and 2_liquidate_by_endDate)
- Initializer scripts: they are executed before running the integration tests. They are used to setup up configuration (like required submissions, and others).
- Init config: used in the initializer scripts.

Unit tests passing
image

Integration tests passing
image

Executing Initializers
image


contracts/util/ZeroCollateralCommon.sol
```diff
@@ -78,6 +78,7 @@ library ZeroCollateralCommon {
         uint256 lastCollateralIn;
         uint256 principalOwed;
         uint256 interestOwed;
+        uint256 borrowedAmount;
```
hensha256, line 81: This is the same as `principalOwed`?

salazarguille: Yes, when the user takes out a loan, the borrowed amount, and principalOwed are set. But after the first repayment, we don't know the original borrowed amount (the principalOwed is modified).
So, then, we could aggregate data based on all the borrowed amounts.

test-integration/liquidate-loan/1_liquidate_by_collateral.js
```diff
@@ -0,0 +1,141 @@
...
+  
+  console.log(`Liquidating loan id ${lastLoanID}...`);
+  const initialLiquidatorTokenBalance = await token.balanceOf(liquidatorTxConfig.from);
+  const initialTotalCollateral = await loansInstance.totalCollateral();
+  const liquidateLoanResult = await loansInstance.liquidateLoan(lastLoanID, liquidatorTxConfig);
+  const liquidateEthPrice = await settingsInstance.liquidateEthPrice();
+  const loanPrinter = new LoanInfoPrinter(web3, loanInfo, { tokenName, decimals });
+  const tokensPaymentIn = loanPrinter.getTotalTokensPaymentInLiquidation(finalOraclePrice, liquidateEthPrice);
+  loans
+    .loanLiquidated(liquidateLoanResult)
```
mark1323, line 125: Still trying to understand whether any (local) DAI/USDC is sent to the liquidation process in exchange for the collateral?

test/utils/loan-terms-helper.js
```diff
@@ -0,0 +1,69 @@
...
+        };
+    */
+    createMultipleSignedLoanTermsResponses: async (web3, loanTermsRequest, signers, loanResponseInfoTemplate, nonces) => {
+        assert(signers && signers.length > 0, 'Requires at least one signer.');
+        const signedLoanTermsResponses = [];
+        for (const signer of signers) {
+            const loanResponseInfo = {
+                ...loanResponseInfoTemplate,
+                signer: signer,
+                signerNonce: nonces.newNonce(signer),
```
mark1323, line 62: It doesn't really matter for this test, I think, but it seems that nonces should be on a per-signer basis.

scripts/lendingPool/deposit.js
```diff
@@ -13,7 +11,7 @@ const processArgs = new ProcessArgs();
 /** Default Parameters: */
 const defaultTokenName = 'USDC';
 const defaultSenderIndex = 0;
-const defaultDepositAmount = 3000000;
+const defaultDepositAmount = '3000000000000';
```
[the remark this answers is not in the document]
salazarguille: Done. It uses a util function.

contracts/base/Loans.sol
```diff
@@ -424,11 +426,11 @@ contract Loans is Base, LoansInterface {
     function _payLoan(uint256 loanID, uint256 toPay) internal {
         if (toPay > loans[loanID].principalOwed) {
             uint256 leftToPay = toPay;
-            leftToPay -= loans[loanID].principalOwed;
+            leftToPay = leftToPay.sub(loans[loanID].principalOwed);
```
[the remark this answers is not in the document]
salazarguille: Yes, it is just in case we (or any other dev) change it in the future. So, we avoid any potential issues in the future.

PR #57, opened 2020-06-19
Fix Oracle Issues

salazarguille:
It contains:

- A new inverse price implementation.
- Configuration for token and response decimals (based on Chainlink team).
- New getCollateralInfo view function.
- New unit tests to verify collateral using new oracle impl.

image

### DIFF
--- a/contracts/base/LoansBase.sol
+++ b/contracts/base/LoansBase.sol
@@ -134,29 +134,6 @@ contract LoansBase is Base {
         _emitCollateralWithdrawnEvent(loanID, msg.sender, withdrawalAmount);
     }
 
-    function getCollateralInfo(uint256 loanID)
-        external
-        view
-        returns (
-            uint256 collateral,
-            uint256 collateralNeededLendingTokens,
-            uint256 collateralNeededCollateralTokens,
-            bool requireCollateral,
-            uint256 oraclePrice
-        )
-    {
-        collateral = loans[loanID].collateral; // Collateral Tokens (ETH, LINK).
-        (
-            collateralNeededLendingTokens,
-            collateralNeededCollateralTokens
-        ) = _getCollateralInfo(
-            _getTotalOwed(loanID),
-            loans[loanID].loanTerms.collateralRatio
-        );
-        requireCollateral = collateralNeededCollateralTokens > collateral;
-        oraclePrice = uint256(priceOracle.getLatestAnswer());
-    }
-
     /**
      * @notice Take out a loan
      *
@@ -170,7 +147,7 @@ contract LoansBase is Base {
         isInitialized()
         whenNotPaused()
         whenLendingPoolNotPaused(address(lendingPool))
-        nonReentrant() // TODO Should it be for TokenLoans?
+        nonReentrant()
         isBorrower(loans[loanID].loanTerms.borrower)
     {
         require(
@@ -194,14 +171,15 @@ contract LoansBase is Base {
             .div(DAYS_PER_YEAR_4DP);
 
         // check that enough collateral has been provided for this loan
-        uint256 collateralNeededToken = _getCollateralNeededInTokens(
-            _getTotalOwed(loanID),
-            loans[loanID].loanTerms.collateralRatio
-        );
-        uint256 collateralNeededWei = _convertTokenToWei(collateralNeededToken);
+        (
+            , // Loan Collateral
+            , // Collateral Needed (in Lending Tokens)
+            , // Collateral Needed (In Collateral)
+            bool requireCollateral
+        ) = _getCollateralInfo(loanID);
 
         require(
-            loans[loanID].collateral >= collateralNeededWei,
+            !requireCollateral,
             "MORE_COLLATERAL_REQUIRED"
         );
 
@@ -230,7 +208,7 @@ contract LoansBase is Base {
         isInitialized()
         whenNotPaused()
         whenLendingPoolNotPaused(address(lendingPool))
-        nonReentrant() // TODO Should it be for TokenLoans?
+        nonReentrant()
     {
         // calculate the actual amount to repay
         uint256 toPay = amount;
@@ -281,20 +259,18 @@ contract LoansBase is Base {
         whenLendingPoolNotPaused(address(lendingPool))
         nonReentrant()
     {
-        // TODO Validate allowance amount.
         // calculate the amount of collateral the loan needs in tokens
-        uint256 collateralNeededToken = _getCollateralNeededInTokens(
-            _getTotalOwed(loanID),
-            loans[loanID].loanTerms.collateralRatio
-        );
-        uint256 collateralNeededWei = _convertTokenToWei(collateralNeededToken);
+        (
+            uint256 loanCollateral,
+            , // Collateral Needed (in Lending Tokens)
+            uint256 collateralNeededWei,
+            // Require Collateral
+        ) = _getCollateralInfo(loanID);
 
         // calculate when the loan should end
         uint256 loanEndTime = loans[loanID].loanStartTime.add(
             loans[loanID].loanTerms.duration
         );
-
-        uint256 loanCollateral = loans[loanID].collateral;
 
         // to liquidate it must be undercollateralised, or expired
         require(
@@ -317,6 +293,19 @@ contract LoansBase is Base {
         lendingPool.liquidationPayment(tokenPayment, msg.sender);
 
         _emitLoanLiquidatedEvent(loanID, msg.sender, loanCollateral, tokenPayment);
+    }
+
+    function getCollateralInfo(uint256 loanID)
+        external
+        view
+        returns (
+            uint256 collateral,
+            uint256 collateralNeededLendingTokens,
+            uint256 collateralNeededCollateralTokens,
+            bool requireCollateral
+        )
+    {
+        return _getCollateralInfo(loanID);
     }
 
     /** Internal Functions */
@@ -346,7 +335,28 @@ contract LoansBase is Base {
         uint256 tokensIn
     ) internal;
 
-    function _getCollateralInfo(uint256 totalOwed, uint256 collateralRatio)
+    function _getCollateralInfo(uint256 loanID)
+        internal
+        view
+        returns (
+            uint256 collateral,
+            uint256 collateralNeededLendingTokens,
+            uint256 collateralNeededCollateralTokens,
+            bool requireCollateral
+        )
+    {
+        collateral = loans[loanID].collateral;
+        (
+            collateralNeededLendingTokens,
+            collateralNeededCollateralTokens
+        ) = _getCollateralNeededInfo(
+            _getTotalOwed(loanID),
+            loans[loanID].loanTerms.collateralRatio
+        );
+        requireCollateral = collateralNeededCollateralTokens > collateral;
+    }
+
+    function _getCollateralNeededInfo(uint256 totalOwed, uint256 collateralRatio)
         internal
         view
         returns (
--- a/contracts/base/LoansBase.sol
+++ b/contracts/base/LoansBase.sol
@@ -171,9 +171,9 @@ contract LoansBase is Base {
             .div(DAYS_PER_YEAR_4DP);
 
         // check that enough collateral has been provided for this loan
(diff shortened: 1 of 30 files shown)


contracts/providers/chainlink/ChainlinkPairAggregator.sol
```diff
@@ -19,15 +19,32 @@ import "@chainlink/contracts/src/v0.5/interfaces/AggregatorInterface.sol";
...
+
     AggregatorInterface public aggregator;
+    uint8 public tokenDecimals;
+    uint8 public responseDecimals;
 
-    constructor(address aggregatorAddress) public {
-        require(aggregatorAddress != address(0x0), "Aggregator address is required.");
+    constructor(address aggregatorAddress, uint8 tokenDecimalsValue, uint8 responseDecimalsValue) public {
+        require(aggregatorAddress != address(0x0), "PROVIDE_AGGREGATOR_ADDRESS");
+        require(tokenDecimalsValue > 0, "PROVIDE_VALID_TOKEN_DECIMALS");
```
hensha256, line 31: Tokens can have 0 decimals - it's rare but its not impossible. I dont think we should rule it out

contracts/providers/chainlink/ChainlinkPairAggregator.sol
```diff
@@ -19,15 +19,32 @@ import "@chainlink/contracts/src/v0.5/interfaces/AggregatorInterface.sol";
...
+        require(responseDecimalsValue > 0, "PROVIDE_VALID_RESPONSE_DECIMALS");
```
hensha256, line 32: Is it guaranteed that chainlink never has a 0 value here?

contracts/providers/chainlink/ChainlinkPairAggregator.sol
```diff
@@ -38,15 +55,34 @@ contract ChainlinkPairAggregator is PairAggregatorInterface {
...
-        require(roundsBack <= latest, "Not enough history");
-        return aggregator.getAnswer(latest - roundsBack);
+        require(roundsBack <= latest, "NOT_ENOUGH_HISTORY");
+        return aggregator.getTimestamp(latest - roundsBack);
     }
 
-    function getPreviousTimestamp(uint256 roundsBack) external view returns (uint256) {
+    /** Internal Functions */
+
+    function _getLatestAnswer() internal view returns (int256) {
```
hensha256, line 66: Why is this function factored out? It's only used once

contracts/providers/chainlink/ChainlinkPairAggregator.sol
```diff
@@ -38,15 +55,34 @@ contract ChainlinkPairAggregator is PairAggregatorInterface {
...
+
+    function _getPreviousAnswer(uint256 roundsBack) internal view returns (int256) {
         uint256 latest = aggregator.latestRound();
-        require(roundsBack <= latest, "Not enough history");
-        return aggregator.getTimestamp(latest - roundsBack);
+        require(roundsBack <= latest, "NOT_ENOUGH_HISTORY");
+        return aggregator.getAnswer(latest - roundsBack);
+    }
+
+    function _normalizeDecimals(int256 value) internal view returns (int256) {
```
hensha256, line 76: Please can you explain the purpose of this function? Why is it that we need the answer to have at least tokenDecimals and maybe more?

contracts/providers/chainlink/ChainlinkPairAggregator.sol
salazarguille: Correct. It's due to I had another impl before. 
Done

contracts/providers/chainlink/ChainlinkPairAggregator.sol
salazarguille: The values returned from Chainlink have different decimals (18,  8, others), and don't match with the token decimals. For example, the oracle LINK/USD has 8 decimals but the Link token has 18; but in case DAI/ETH, it has 18 decimals.
We use it to get the oracle price in "token decimals". So then, for example, we can get the inverse rate.
I have created unit tests where we mock the oracle price (based on real data from Ropste) and calculate the collateral info: LoansBaseGetCollateralInfoTest

For example, one of the unit tests is:
Price LINK/USD (Chainlink): 414000000 (8 decimals), but we need to use the inverse rate (USD/LINK).

contracts/providers/chainlink/ChainlinkPairAggregator.sol
salazarguille: They have mentioned today, the response is similar to token decimals. So it might be zero.
I removed the validation.

contracts/providers/chainlink/ChainlinkPairAggregator.sol
hensha256: > We use it to get the oracle price in "token decimals".
If chainlink has 18 decimals and the token has 12, then the function gives the result 18 decimals - which is the same number of decimals as the chainlink oracle? How does the code then deal with that?

salazarguille: The oracle price is only used in the LoansBase contracts (LoansBase._convertWeiToToken and LoansBase._convertTokenToWei). It deals in the same way that it does for the other scenarios.

There are unit tests for that scenario in the unit test test/base/LoansBaseGetCollateralInfoTest.js
- _4_usdc_eth_response_18_token_6_repay_0

This info is printed using the LoansBase.getCollateralInfo function (it uses LoansBase._convertTokenToWei internally)
In Lending Tokens: 50000000 USDC (6 decimals)
In Coll Tokens:    215645250000000000 WEI = 0.21564525 ETH
Oracle Price:      4312905000000000 => 1 USDC = 0.004312905 ETH

In that case, the "In Coll Tokens" is calculated 0.004312905 * 50 USDC = 0.21564525 ETH

The _convertXToY functions use a "whole lending token to calculate the output.

- _5_usdc_eth_response_18_token_6_repay_50
- _6_usdc_eth_response_18_token_6_repay_100

hensha256: Okay here is a scenario that I think it is wrong... Maybe I am misunderstanding though
- We have a market with USDC (6 decimals) as lending token
- SLINK (made up token... 8 decimals) as collateral token.
- The price oracle gives the price for 1 whole USDC in SLINK.
- The price oracle has 12 decimals

Let's say the price in for 1 USDC in whole SLINK is 3, so the price for 1 whole USDC in SLINK units is 300,000,000 (8 decimals). The Oracle uses 12 decimals and so has the price 3,000,000,000,000.

`LoansBase._convertTokenToWei` converts USDC units, to their unit SLINK cost

So lets say we call `LoansBase._convertTokenToWei(4000000)` i.e. 4 USDC units to SLINK units
```
uint256 aWholeLendingToken = 10**6
 // because tokenDecimals >= responseDecimals is not true, the next line does not alter the oracle's decimals
uint256 oneLendingTokenPriceWei = 3,000,000,000,000
uint256 tokenValue = 4000000.mul(3,000,000,000,000).div(10**6);
                                 = 12,000,000,000,000
```
So this function says that 4 USDC units is 12 trillion SLINK units, which is 120,000 whole SLINK. However we know that it should be 12 SLINK for 4 USDC.

The reason for this difference in output is because the oracle has 4 more decimals than the token.

salazarguille: Great catch! I have added 30 unit tests (total) to cover your scenario and all the possible ones.

contracts/base/LoansBase.sol
```diff
@@ -193,16 +171,9 @@ contract LoansBase is Base {
...
-            loans[loanID].loanTerms.collateralRatio
-        );
-        uint256 collateralNeededWei = _convertTokenToWei(collateralNeededToken);
+        (, , , bool requireCollateral) = _getCollateralInfo(loanID);
 
-        require(
-            loans[loanID].collateral >= collateralNeededWei,
-            "MORE_COLLATERAL_REQUIRED"
-        );
+        require(!requireCollateral, "MORE_COLLATERAL_REQUIRED");
```
hensha256, line 176: I think `requireCollateral` is a bad variable name, because `require(!requireCollateral` reads 
"require not require collateral" which is a weird statement

I think that "moreCollateralRequired" makes more sense (like your revert message)

contracts/base/LoansBase.sol
```diff
@@ -345,7 +324,28 @@ contract LoansBase is Base {
         uint256 tokensIn
     ) internal;
 
-    function _getCollateralInfo(uint256 totalOwed, uint256 collateralRatio)
+    function _getCollateralInfo(uint256 loanID)
```
hensha256, line 327: Does this function make calls less efficient? The places that call `_getCollateralInfo` seem to only use 1/2 of the outputs it gives and so don't need the other values to be calculated.

Maybe this is an improvement for later

contracts/base/LoansBase.sol
```diff
@@ -280,21 +251,16 @@ contract LoansBase is Base {
         whenLendingPoolNotPaused(address(lendingPool))
         nonReentrant()
     {
-        // TODO Validate allowance amount.
         // calculate the amount of collateral the loan needs in tokens
-        uint256 collateralNeededToken = _getCollateralNeededInTokens(
-            _getTotalOwed(loanID),
-            loans[loanID].loanTerms.collateralRatio
+        (uint256 loanCollateral, , uint256 collateralNeededWei, ) = _getCollateralInfo(
```
hensha256, line 255: I think we need to stop using the word 'wei' in this contract when it's not necessarily ETH.

contracts/base/LoansBase.sol
```diff
@@ -280,21 +251,16 @@ contract LoansBase is Base {
...
         // calculate when the loan should end
         uint256 loanEndTime = loans[loanID].loanStartTime.add(
             loans[loanID].loanTerms.duration
         );
 
-        uint256 loanCollateral = loans[loanID].collateral;
-
         // to liquidate it must be undercollateralised, or expired
         require(
             collateralNeededWei > loanCollateral || loanEndTime < now,
```
hensha256, line 266: The value of `collateralNeededWei > loanCollateral` is just the variable `requireCollateral` that was calculated by `_getCollateralInfo` on line 255.

contracts/providers/chainlink/ChainlinkPairAggregator.sol
```diff
@@ -38,15 +53,31 @@ contract ChainlinkPairAggregator is PairAggregatorInterface {
...
+        if( collateralDecimals >= responseDecimals) {
+            uint8 pendingDecimals = collateralDecimals - responseDecimals;
+            return value * int256(TEN ** pendingDecimals);
+        } else {
+            uint8 pendingDecimals = responseDecimals - collateralDecimals;
+            return value / int256(TEN ** pendingDecimals);
+        }
+    }
+
+    function _normalizeResponse(int256 response) internal view returns (int256) {
```
hensha256, line 80: What's the purpose of this function? Why not just call `_normalizeDecimals` directly instead?

hensha256: I see its so it works with the inverseaggregator...
why not just define `_normalizeResponse` as the above decimal work, and then in the inverse aggregator defined `_normalizeResponse` as the inversion maths? It seems weird to just have a few function calls that purely call another function

contracts/base/LoansBase.sol
salazarguille: The output variables that are being used require the other output variables to be calculated internally. The difference is now we are returning all the variables.
The function was created using the previous logic. It just put together two functions (_getCollateralNeededInTokens and _convertTokenToWei)
For sure, we can improve it later.

contracts/base/LoansBase.sol
salazarguille: I will change it to "collateralNeededInCollTokens".

contracts/providers/chainlink/ChainlinkPairAggregator.sol
salazarguille: Okay, so I removed the _normalizeDecimals function (moved logic to _normalizeResponse). The inverse aggregator already has the _normalizeResponse as you mentioned.

PR #60, opened 2020-06-23
Adding Docs for Truffle Scripts

salazarguille:
It includes:

- Docs to detail how to execute truffle scripts.


docs/scripts.md
```diff
@@ -1,26 +1,66 @@
 # Truffle Scripts
```
ark4n631, line 1: shouldn't the filename would be `docs/truffle-scripts.md` just to be consistent with the other docs

ark4n631:
Just as i commented to you in slack, i think:
image
because this project doesn't work by default on nodeJS LTS `v12.x` so you are forced to use an older nodeJS version, that should be warned before you `yarn install` and it seems hidden in a note below check truffle (when all should be already installed):

image

docs/scripts.md
salazarguille: Done

salazarguille:
Great, I pushed the changes.

PR #62, opened 2020-06-24
New Contracts Conf

salazarguille:
It includes:

- New contract addresses on ropsten with the last version.
- New ganache snapshot for integration tests.
- Small fixes.


contracts/mock/providers/chainlink/InverseChainlinkPairAggregatorMock.sol
```diff
@@ -28,7 +28,7 @@ contract InverseChainlinkPairAggregatorMock is InverseChainlinkPairAggregator {
 
     /** Internal Functions */
     
-    function _normalizeDecimals(int256 value) internal view returns (int256) {
+    function _normalizeDecimals(int256 value) internal pure returns (int256) {
```
hensha256, line 31: This function can be removed completely - it's never used anywhere anymore

salazarguille: You're right. I removed this contract. Now we are using the real contract in the test.

PR #68, opened 2020-07-13
Audit: Best practices Fixes

salazarguille:
It includes fixes for the "Adherence to Best Practices" issues found in the audit process.

image

### DIFF
--- a/contracts/base/Lenders.sol
+++ b/contracts/base/Lenders.sol
@@ -122,6 +122,7 @@ contract Lenders is Base, LendersInterface {
         isInitialized()
         returns (uint256)
     {
+        require(amount > 0, "CANNOT_WITHDRAW_ZERO");
         require(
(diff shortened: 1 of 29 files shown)


contracts/base/Lenders.sol
```diff
@@ -122,22 +122,18 @@ contract Lenders is Base, LendersInterface {
...
-
-        if (withdrawableInterest == 0 || amountToWithdraw == 0) return 0;
-
-        if (withdrawableInterest < amountToWithdraw) {
-            amountToWithdraw = withdrawableInterest;
-        }
-
-        accruedInterest[recipient].totalNotWithdrawn = withdrawableInterest.sub(
-            amountToWithdraw
+        require(
```
alex-murashkin, line 125: Would it also make sense to `require(...)` the `amount` to be non-zero?

salazarguille: Fixed!

PR #70, opened 2020-07-16
Fix High and Medium Issues

salazarguille:
It includes fixes for the "High and Medium" issues found in the audit process

image

### DIFF
--- a/contracts/base/TokenCollateralLoans.sol
+++ b/contracts/base/TokenCollateralLoans.sol
@@ -24,6 +24,8 @@ import "./LoansBase.sol";
 contract TokenCollateralLoans is LoansBase {
     /** Constants */
 
+    bool internal constant IS_TRANSFER = true;
+
     /** Properties */
 
     /** Modifiers */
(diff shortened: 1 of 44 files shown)


contracts/base/TokenCollateralLoans.sol
```diff
@@ -147,6 +147,24 @@ contract TokenCollateralLoans is LoansBase {
         collateralTokenTransferFrom(msg.sender, amount);
     }
 
+    function _requireExpectedBalance(uint256 initialBalance, uint256 expectedAmount)
+        internal
+        view
+    {
+        uint256 finalBalance = ERC20Detailed(collateralToken).balanceOf(address(this));
+        if (finalBalance > initialBalance) {
```
alex-murashkin, line 155: I think it would be better to specify an additional function parameter indicating the direction of transfer, and use that for the assertions. Using the condition `if (finalBalance > initialBalance)` reads a little weird to me.

salazarguille: Fixed, now it receives isTransfer param. Also, I added some unit tests.

PR #71, opened 2020-07-16
Fix Low and Info Issues

salazarguille:
It includes fixes for the "Low and info" issues found in the audit process

image

### DIFF
--- a/contracts/base/Settings.sol
+++ b/contracts/base/Settings.sol
@@ -31,7 +31,7 @@ contract Settings is Pausable, SettingsInterface {
 
     /** Constants */
     /**
-        The maximum tolerance is a porcentage with 2 decimals.
+        The maximum tolerance is a percentage with 2 decimals.
         Examples:
             350 => 3.5%
             4000 => 40.00%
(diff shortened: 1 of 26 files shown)


contracts/base/Settings.sol
```diff
@@ -30,6 +30,16 @@ contract Settings is Pausable, SettingsInterface {
     using AddressLib for address;
 
     /** Constants */
+    /**
+        The maximum tolerance is a porcentage with 2 decimals.
```
alex-murashkin, line 34: a typo: `percentage`.

salazarguille: Fixed

contracts/base/TokenCollateralLoans.sol
```diff
@@ -147,6 +147,24 @@ contract TokenCollateralLoans is LoansBase {
         collateralTokenTransferFrom(msg.sender, amount);
     }
 
+    function _requireExpectedBalance(uint256 initialBalance, uint256 expectedAmount)
```
alex-murashkin, line 150: It seems like it overlaps with changes in #70 , let's make sure that we use the updated function here, after that PR is merged.

PR #72, opened 2020-07-16
Standardize Error Messages

salazarguille:
It includes:

- Changes in some error messages to standardize them using uppercase and underscores, and less than 32 chars:

Examples:
MY_ERROR_MESSAGE

image


salazarguille:
It was implemented in this PR #88

PR #82, opened 2020-07-23
Features/check pre transfer balance

sypeer:
Added "require" validations before a contract executes:
- a transfer function or 
- a transferFrom function

This is to verify whether the contract has enough balance or allowance and display user-friendly error messages to the final user instead, of non-readable error messages 

Screen Shot 2020-07-24 at 6 31 12 PM


contracts/base/LendingPool.sol
```diff
@@ -267,6 +273,13 @@ contract LendingPool is Base, LendingPoolInterface {
         @dev It throws a require error if 'transferFrom' invocation fails.
      */
     function tokenTransferFrom(address from, uint256 amount) private {
+        // The first param is the  owner and second the spender. That's why the unit tests were failing. :D
+        // If I could putting a slapping my head emoji here I would lol :S
+        uint256 allowance = lendingToken.allowance(address(this), from);
```
salazarguille, line 278: I think the "from" should be first because it is the owner. Then the spender, in this case, the contract address(this).

test/base/LendingPoolTransferTest.js
```diff
@@ -0,0 +1,98 @@
...
+            const mintResponse = compoundFails ? 1 : 0
+            const encodeCompMint = compoundInterfaceEncoder.encodeMint();
+            await cTokenInstance.givenMethodReturnUint(encodeCompMint, mintResponse);
+
+            const encodeAllowance = erc20InterfaceEncoder.encodeAllowance();
+            await daiInstance.givenMethodReturnUint(encodeAllowance, allowance);
+
+            try {
+                // Invocation
+                const result = await instance.repay(amountToRepay, borrower, { from: sender });
```
salazarguille, line 82: This unit test is testing the  "repay"  function. So, the filename and name (see line #17) should match with the function it is testing.

contracts/base/LendingPool.sol
sypeer: Fixed

sypeer: Fixed

test/base/LendingPoolTransferTest.js
sypeer: removed

PR #83, opened 2020-07-23
[PROTOCOL-78] Audit: Increase Code Coverage

sypeer:
It includes:

- New unit tests to increase code coverage.

image


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
salazarguille, line 1: I think we don't need this file. Each dev uses a different configuration for the VSCode (or use another IDE)

test/base/EtherCollateralLoansGetBorrowerLoansTest.js
```diff
@@ -107,6 +107,10 @@ contract('EtherCollateralLoansCreateLoanWithTermsTest', function (accounts) {
 
             const loan = await instance.loans.call(mockLoanIDCounter)
             
+            const borrowedLoans = await instance.getBorrowerLoans(borrowerAddress);
```
salazarguille, line 110: I think if you want to create unit tests for the getBorrowerLoans function, it is better if you create a new unit test file  (EtherCollateralLoansGetBorrowerLoansTest).

We don't want to change a test for another function. Also, remember the filename and the name at the top of the file (contract('EtherCollateralLoansCreateLoanWithTermsTest') )have to match.

test/base/LoanTermsConsensusProcessRequestTest.js
```diff
@@ -167,5 +170,79 @@ contract('LoanTermsConsensusProcessRequestTest', function (accounts) {
                 assert.equal(error.reason, expectedErrorMessage);
             }
         })
-    })
+    });
+
+    withData({
+        _9_nonce_taken: [
+            1, 320, [responseOne], true, 'REQUEST_NONCE_TAKEN'
```
salazarguille, line 177: I think this scenario is already covered on the branch develop here https://github.com/teller-protocol/teller-protocol-v1/blob/develop/test/base/LoanTermsConsensusProcessRequestTest.js#L108
Please review it is covering the same scenario.

salazarguille: Fixed

test/base/EtherCollateralLoansGetBorrowerLoansTest.js
salazarguille: Fixed

.vscode/settings.json
salazarguille: Fixed

.solcover.js
```diff
@@ -5,5 +5,6 @@ module.exports = {
     compileCommand: 'node --max-old-space-size=4096 ../node_modules/.bin/truffle compile --network coverage',
     skipFiles: [
         'mock/',
+        'providers/openzeppelin/SignedSafeMath.sol' // It is not included due to it is a OpenZeppelin contract.
```
alex-murashkin, line 8: 👍

PR #84, opened 2020-07-23
[PROTOCOL-81] Adding Docs In Contracts II

sypeer:
It includes docs for:

- LoansInterface and Implementations (Ether and Token)
- LoansBase
- LoanTermsConsensusInterface and Implementation (LoanTermConsensus)

In each contract:
- Added docs in external / public / internal functions, constructor and events. (using @ annotations).
- The contract has a header describing the purpose and @author develop@teller.finance

### DIFF
--- a/contracts/base/LoanTermsConsensus.sol
+++ b/contracts/base/LoanTermsConsensus.sol
@@ -19,8 +19,8 @@ contract LoanTermsConsensus is Consensus, LoanTermsConsensusInterface {
 
     /**
         @notice Processes the loan request
-        @param ZeroCollateralCommon.LoanRequest request Struct of the protocol loan request
-        @param ZeroCollateralCommon.LoanResponses request List of structs of the protocol loan responses
+        @param request Struct of the protocol loan request
+        @param responses List of structs of the protocol loan responses
         @return uint256 Interest rate
         @return uint256 Collateral ratio
         @return uint256 Maximum loan amount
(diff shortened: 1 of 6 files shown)


contracts/base/LoanTermsConsensus.sol
```diff
@@ -22,11 +7,24 @@ import "./Consensus.sol";
...
 contract LoanTermsConsensus is Consensus, LoanTermsConsensusInterface {
+    /* Mappings */
     mapping(address => mapping(uint256 => ZeroCollateralCommon.AccruedLoanTerms)) public termSubmissions;
     mapping(address => mapping(uint256 => bool)) public requestNonceTaken;
 
+    /**
+        @notice Processes the loan request
+        @param ZeroCollateralCommon.LoanRequest request Struct of the protocol loan request
+        @param ZeroCollateralCommon.LoanResponses request List of structs of the protocol loan responses
+        @return uint256 Interest rate
```
salazarguille, line 24: It should be @return interestRate Interest rate and the same for the other return params

contracts/base/LoanTermsConsensus.sol
```diff
@@ -71,6 +69,12 @@ contract LoanTermsConsensus is Consensus, LoanTermsConsensusInterface {
         );
     }
 
+    /**
+        @notice Processes the loan response
+        @param ZeroCollateralCommon.LoanRequest request Struct of the protocol loan request
```
salazarguille, line 74: It should be @param request struct of the protocol loan request

contracts/base/LoanTermsConsensus.sol
sypeer: fixed

contracts/base/LoanTermsConsensus.sol
sypeer: fixed

contracts/base/LoansBase.sol
```diff
@@ -305,10 +330,23 @@ contract LoansBase is LoansInterface, Base {
     }
 
     /** Internal Functions */
-
+    /**
+        @notice Payouts the collateral for a loan
```
alex-murashkin, line 334: I think it’s meant to be “pays out”

sypeer: Fixed, thank you

PR #99, opened 2020-08-01
[PROTOCOL-71] Node components version tracking feature

Juan-cc:
It is:
 - Settings contract added capability to manage a list of Node Components and their versions.

It includes:
- New functions, events and mapping on Settings.sol and SettingsInterface.sol.
- Default Node components version configuration.
- New unit tests.
- Settings contract deployment files changed to add default component versions.

image


contracts/base/Settings.sol
```diff
@@ -351,6 +359,66 @@ contract Settings is Pausable, SettingsInterface {
...
+        @param componentName name of the component to be added.
+        @param minVersion minimum component version supported.
+     */
+    function addNewComponent(bytes32 componentName, uint256 minVersion)
+        external
+        onlyPauser()
+        whenNotPaused()
+    {
+        require(minVersion > 0, "INVALID_COMPONENT_MINIMUM_VERSION");
+        require(componentName[0] != 0, "MISSING_COMPONENT_NAME");
```
salazarguille, line 373: I think it can be checked in this way:

require(componentName != "", "MISSING_COMPONENT_NAME")

Also, I think the error message could be "COMPONENT_NAME_MUST_BE_PROVIDED" (or similar) (just to be consistent with the rest of the errors). Please, verify the error message length is < 32 chars.

contracts/base/Settings.sol
```diff
@@ -351,6 +359,66 @@ contract Settings is Pausable, SettingsInterface {
...
+    function updateComponentVersion(bytes32 componentName, uint256 newMinVersion)
+        external
+        onlyPauser()
+        whenNotPaused()
+    {
+        require(newMinVersion > 0, "INVALID_COMPONENT_MINIMUM_VERSION");
+        require(componentName[0] != 0, "MISSING_COMPONENT_NAME");
+        require(minimumNodeRequirements[componentName] > 0, "COMPONENT_NOT_FOUND");
+        require(
+            newMinVersion != minimumNodeRequirements[componentName],
```
salazarguille, line 409: Do we need to validate newVersion > currentVersion?

config/consts.js
```diff
@@ -9,4 +9,9 @@ module.exports = {
         This is used as a default amount to configure max lending amount.
      */
     DEFAULT_MAX_AMOUNT: 1000,
+
+    /**
+     * Represents a fake Node Component version.
+     */
+    DUMMY_NODE_COMPONENT_VERSION: 123,
```
salazarguille, line 16: I would set the initial version directly (not a dummy value) describing the format. Maybe 1.02.03.04 => 1020304?

migrations/utils/initSettings.js
```diff
@@ -40,4 +42,68 @@ module.exports = async function(
...
+}
+
+/**
+ * We add initial Node Components version to Settings contract to start the platform.
+ * 
+ * @param settingsInstance a Settings contract instance. 
+ * @param web3 web3 library reference.
+ */
+function initNodeComponents(settingsInstance, nodeComponentsVersions, web3) {
+    const WEB2 = 'WEB2';
```
salazarguille, line 54: I would create a JS file with the constants (maybe test/utils/componentNames.js). Similar to test/utils/chains.js

contracts/interfaces/SettingsInterface.sol
```diff
@@ -210,6 +248,21 @@ interface SettingsInterface {
...
+    function addNewComponent(bytes32 componentName, uint256 minVersion) external;
+
+    /**
+        @notice Set a new version for a Node Component.
+        @param componentName name of the component
+        @param newMinVersion minimum component version supported
+     */
+    function updateComponentVersion(bytes32 componentName, uint256 newMinVersion)
+        external;
+
```
salazarguille, line 265: I think the remove function is missing.

contracts/base/Settings.sol
```diff
@@ -351,6 +359,66 @@ contract Settings is Pausable, SettingsInterface {
...
+    {
+        require(componentName[0] != 0, "MISSING_COMPONENT_NAME");
+        require(minimumNodeRequirements[componentName] > 0, "COMPONENT_NOT_FOUND");
+        uint256 previousVersion = minimumNodeRequirements[componentName];
+        delete minimumNodeRequirements[componentName];
+        emit NodeComponentRemoved(msg.sender, componentName, previousVersion);
+    }
+
+    /**
+        @notice Set a new version for a Node Component.
```
salazarguille, line 396: Should we have a getter (view) function to get the current version for a given component (and in the interface too)?

contracts/base/Settings.sol
Juan-cc: Perfect, I already applied this suggestion.

contracts/base/Settings.sol
Juan-cc: Suggestion added to validations and a new comment will be shared on the Jira ticket as this wouldn't let us go back on any version in case of an error (typo).

contracts/interfaces/SettingsInterface.sol
Juan-cc: I added it, thanks for catching this

config/consts.js
Juan-cc: I set 1000000 representing 1.00.00.00 but I am confirming this with Esteban which is going to be the initial version of each component.

migrations/utils/initSettings.js
Juan-cc: Extracted constants into a newly created constants file with components' names.

contracts/interfaces/SettingsInterface.sol
Juan-cc: Correct, I added this function in the Interface too.

contracts/base/Settings.sol
Juan-cc: Getter created.

contracts/base/Settings.sol
Juan-cc: Suggestions implemented

contracts/base/Settings.sol
salazarguille: Okay, in case of an error, I think we might rollback the change in the node component (but increase the version to a new one instead of rollbacking the version too). So, the version is always incremental.

contracts/base/Settings.sol
```diff
@@ -86,6 +86,13 @@ contract Settings is Pausable, SettingsInterface {
      */
     mapping(address => bool) public lendingPoolPaused;
 
+    /**
+        @notice Contains minimum version for each node component.
+
+        i.e.: "web2" => "1234500" represents "web2" => "01.23.45.00"
+     */
+    mapping(bytes32 => uint32) public minimumNodeRequirements;
```
salazarguille, line 94: I would change minimumNodeRequirements to componentVersions. So, it matches with the function (update/create/remove) names.

contracts/base/Settings.sol
```diff
@@ -363,6 +370,75 @@ contract Settings is Pausable, SettingsInterface {
...
+        @param componentName name of the component to be added.
+        @param minVersion minimum component version supported.
+     */
+    function createComponentVersion(bytes32 componentName, uint32 minVersion)
+        external
+        onlyPauser()
+        whenNotPaused()
+    {
+        require(minVersion > 0, "INVALID_COMPONENT_VERSION");
+        require(componentName[0] != 0, "COMPONENT_NAME_MUST_BE_PROVIDED");
```
salazarguille, line 384: I think componentName[0] != 0 should be componentName != ""

migrations/utils/init_settings/initNodeComponentsSettings.js
```diff
@@ -0,0 +1,23 @@
+const {
```
salazarguille, line 1: I would change the file name to initComponentVersions.js or similar to be consistent with the contract functions.

config/consts.js
```diff
@@ -9,4 +9,12 @@ module.exports = {
         This is used as a default amount to configure max lending amount.
      */
     DEFAULT_MAX_AMOUNT: 1000,
+
+    /**
+     * Represents the initial version of the Node Component.
+     * 
+     * e.g.: 1000000 => represents : 1.00.00.00
+     * e.g.: 10000 => represents : 1.00.00
```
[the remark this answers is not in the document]
Juan-cc: I'll use option B as Esteban mentioned "1.00.00" as initial version

contracts/base/Settings.sol
```diff
@@ -363,6 +370,75 @@ contract Settings is Pausable, SettingsInterface {
...
+        @notice Set a new version for a Node Component.
+        @param componentName name of the component to be modified.
+        @param newVersion minimum component version supported.
+     */
+    function updateComponentVersion(bytes32 componentName, uint32 newVersion)
+        external
+        onlyPauser()
+        whenNotPaused()
+    {
+        require(newVersion > 0, "INVALID_COMPONENT_VERSION");
```
salazarguille, line 425: I think as we are validating "componentVersions[componentName] > 0" and "newVersion > componentVersions[componentName]", then the validation the "newVersion > 0' validation is not required.

contracts/base/Settings.sol
```diff
@@ -363,6 +370,75 @@ contract Settings is Pausable, SettingsInterface {
...
+        require(componentVersions[componentName] == 0, "COMPONENT_ALREADY_EXISTS");
+        componentVersions[componentName] = minVersion;
+        emit ComponentVersionCreated(msg.sender, componentName, minVersion);
+    }
+
+    /**
+        @notice Remove a Node Component from the list.
+        @param componentName name of the component to be removed.
+     */
+    function removeComponentVersion(bytes32 componentName)
```
passabilities, line 394: Should a version be allowed to be removed? Seems like something we should keep a record of on-chain.

migrations/2_deploy.js
```diff
@@ -68,7 +68,7 @@ module.exports = async function(deployer, network, accounts) {
   );
   const settingsInstance = await Settings.deployed();
   await initSettings(
-    settingsInstance,
+    settingsInstance, web3,
```
passabilities, line 71: Should probably go on a new line :sweat_smile:

contracts/base/Settings.sol
Juan-cc: Redundant validation has been removed

PR #100, opened 2020-08-01
[PROTOCOL-33] Create Signature library

sypeer:
It is:
- A reusable Signature library based on functions built in the InterestConsensus and LoanTermsConsensus contracts

It includes:
- New signature library & mock contract for Interest and LoanTerms
- New unit tests 

Screen Shot 2020-08-01 at 1 17 12 PM


test/util/SignatureLibTest.js
```diff
@@ -0,0 +1,176 @@
+// JS Libraries
```
Juan-cc, line 1: I would advice having 1 test file per function. 
e.g.: SignatureLibHashInterestRequestTest.js, SignatureLibHashInterestResponseTest.js, etc
This will keep our test files smaller, easier to handle and modify.

test/util/SignatureLibTest.js
```diff
@@ -0,0 +1,176 @@
...
+    let consensusInstance;
+    let instance;
+
+    beforeEach('Setup for each test', async () => {
+        instance = await SignatureLibMock.new();
+        consensusInstance = await Mock.new();
+    });
+
+    withData({
+        _1_first_test_hashInterestRequest: [chains.mainnet, accounts[2], 234764, 344673177, 34467317723],
```
Juan-cc, line 28: Just a suggestion: As we have a number already in the name and we will split this test file into smaller files, I don't think we need to clarify **_first_test_** or **_second_test_** on each case.

test/util/SignatureLibTest.js
sypeer: Understood, tests have been split now

test/util/SignatureLibTest.js
sypeer: I agree that it was a little unclear. I've renamed the tests to reflect it more accurately

PR #102, opened 2020-08-03
[PROTOCOL-22] Add Setter Function For Oracle Price Reference

salazarguille:
It includes:

- A setter function to update the oracle reference.
- New unit tests.

image

### DIFF
--- a/contracts/base/Base.sol
+++ b/contracts/base/Base.sol
@@ -73,7 +73,7 @@ contract Base is Initializable, ReentrancyGuard {
         @param anAddress account to test.
      */
     modifier whenAllowed(address anAddress) {
-        require(_isAllowed(anAddress), "ADDRESS_ISNT_ALLOWED");
+        require(settings.hasPauserRole(anAddress), "ADDRESS_ISNT_ALLOWED");
         _;
     }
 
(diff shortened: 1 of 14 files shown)


contracts/base/Base.sol
```diff
@@ -67,6 +67,16 @@ contract Base is Initializable, ReentrancyGuard {
...
+    /**
+        @notice Checks whether a given address is allowed (see Settings#hasPauserRole function) or not.
+        @dev It throws a require error if address is not allowed.
+        @param anAddress account to test.
+     */
+    modifier whenAllowed(address anAddress) {
+        require(_isAllowed(anAddress), "ADDRESS_ISNT_ALLOWED");
+        _;
+    }
+
```
passabilities, line 79: seems redundant to have an additional modifier and function that do the same thing as preexisting modifiers.
See other comments as well.

contracts/base/Base.sol
```diff
@@ -103,5 +113,14 @@ contract Base is Initializable, ReentrancyGuard {
...
 
+    /**
+        @notice Tests whether an account has the pauser role.
+        @param anAddress account to test.
+        @return true if account has the pauser role. Otherwise it returns false.
+     */
+    function _isAllowed(address anAddress) internal view returns (bool) {
+        return settings.hasPauserRole(anAddress);
+    }
+
```
passabilities, line 124: seems redundant

contracts/base/Base.sol
salazarguille: Fixed

contracts/base/Base.sol
salazarguille: Fixed

PR #103, opened 2020-08-03
[PROTOCOL-94] Add Collateral Buffer Settings

salazarguille:
It includes:

- New collateral buffer settings.
- New unit tests.
- Changes in unit tests.

image

### DIFF
--- a/contracts/base/Settings.sol
+++ b/contracts/base/Settings.sol
@@ -69,6 +69,10 @@ contract Settings is Pausable, SettingsInterface {
      */
     bytes32 public constant STARTING_BLOCK_NUMBER_SETTING = "StartingBlockNumber";
     /**
+        @notice The setting name for the collateral buffer settings.
+     */
+    bytes32 public constant COLLATERAL_BUFFER_SETTING = "CollateralBuffer";
+    /**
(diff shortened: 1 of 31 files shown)


contracts/base/Settings.sol
```diff
@@ -385,6 +397,28 @@ contract Settings is Pausable, SettingsInterface {
...
+        @notice Sets the collateral buffer value.
+        @param newCollateralBuffer the new collateral buffer value.
+     */
+    function setCollateralBuffer(uint256 newCollateralBuffer)
+        external
+        onlyPauser()
+        whenNotPaused()
+    {
+        require(collateralBuffer != newCollateralBuffer, "NEW_VALUE_REQUIRED");
+        require(newCollateralBuffer > 0, "MUST_PROVIDE_COLLATERAL_BUFFER");
```
Juan-cc, line 410: probably mentioning the collateral value must be a positive number would better clarify the intention of this validation:

MUST_PROVIDE_COLLATERAL_BUFFER  => COLLATERAL_MUST_BE_POSITIVE

This is just a comment.

salazarguille: The Settings contract will be refactored in the next PR. Now it will contain a mapping for the settings

PR #106, opened 2020-08-05
New Deployment on Rinkeby

sypeer:
It is:
- Deployed protocol on Rinkeby TestNet


config/networks/rinkeby/zerocollateral.js
```diff
@@ -0,0 +1,29 @@
+
```
passabilities, line 1: Should probably change this to Teller?

sypeer: Will be done in a subsequent Task

PR #109, opened 2020-08-07
[PROTOCOL-131] Fix Truffle Scripts

salazarguille:
It includes:

- Fixes in truffle scripts.
- Changes in scripts.
- New truffle scripts.


scripts/utils/cli/params.js
```diff
@@ -124,12 +129,34 @@ module.exports.addNewValue = (yargs, defaultParam = NEW_VALUE.default) => {
...
   );
 };
 
+module.exports.addMaxValue = (yargs, defaultParam = MAX_VALUE.default) => {
+  newOption(
+    yargs,
+    MAX_VALUE.name,
+    MAX_VALUE.alias,
+    "string",
+    `Min value to set in the settings. By default: ${defaultParam}`,
```
Juan-cc, line 144: We should update     `Min value to set in the settings. By default: ${defaultParam}`,
with     `Max value to set in the settings. By default: ${defaultParam}`,

PR #112, opened 2020-08-08
[PROTOCOL-37] Calculate Supply to Debt (StD) Ratio

salazarguille:
It includes:

- New contract (library and interface) to accumulate the total amounts, and calculate the supply-to-debt ratio.
- New unit tests.

This is a prototype and we will add the calls to the MarketsState functions in different contracts (Loans.createLoanWithTerms/takeOutLoan/liquidate Lenders.deposit and others). So, the total amounts will be updated dynamically.

image

### DIFF
--- a/contracts/base/MarketsState.sol
+++ b/contracts/base/MarketsState.sol
@@ -1,0 +1,136 @@
+pragma solidity 0.5.17;
+pragma experimental ABIEncoderV2;
+
+// Libraries
+import "@openzeppelin/contracts/math/SafeMath.sol";
+import "@openzeppelin/contracts/access/roles/WhitelistedRole.sol";
+
+// Contracts
+
+// Interfaces
+import "../interfaces/MarketsStateInterface.sol";
+import "../util/MarketStateLib.sol";
+
+
+/**
+    @notice This contract is used to store market data.
+
+    @author develop@teller.finance
+ */
+contract MarketsState is MarketsStateInterface, WhitelistedRole {
+    using SafeMath for uint256;
+    using MarketStateLib for MarketStateLib.MarketState;
+
+    /** Constants */
+
+    /* State Variables */
+
+    /**
+        @notice It maps a lent token => collateral token => Market state.
+        Example:
+            address(DAI) => address(LINK) => MarketState
+            address(DAI) => address(ETH) => MarketState
+     */
+    mapping(address => mapping(address => MarketStateLib.MarketState)) public markets;
+
+    /**
+        @notice It increases the repayment amount for a given market.
+        @param borrowedAsset borrowed asset address.
+        @param collateralAsset collateral asset address.
+        @param amount amount to add.
+     */
+    function increaseRepayment(
+        address borrowedAsset,
+        address collateralAsset,
+        uint256 amount
+    ) external onlyWhitelisted() {
+        markets[borrowedAsset][collateralAsset].increaseRepayment(amount);
+    }
+
+    /**
+        @notice It increases the supply amount for a given market.
+        @param borrowedAsset borrowed asset address.
+        @param collateralAsset collateral asset address.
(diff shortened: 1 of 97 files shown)


test/base/MarketsStateGetMarketTest.js
```diff
@@ -0,0 +1,108 @@
...
+            ], 0, 1, newExpected(2000, 100, 500)
+        ],
+        // (2000 borrow - 1000 repay) / 2000 Supply = 0.5
+        _2_scenario: [
+            [
+                newAmount(1000, actions.Supply, 0, 1),
+                newAmount(1000, actions.Supply, 0, 1),
+                newAmount(500, actions.Borrow, 0, 1),
+                newAmount(500, actions.Repay, 0, 1),
+                newAmount(500, actions.Repay, 0, 1),
```
Juan-cc, line 44: should we be able to repay more than what's borrowed? First repay is ok, but second without borrowed amount should be allowed or rejected?

contracts/base/MarketsState.sol
```diff
@@ -0,0 +1,136 @@
...
+import "../interfaces/MarketsStateInterface.sol";
+import "../util/MarketStateLib.sol";
+
+
+/**
+    @notice This contract is used to store market data.
+
+    @author develop@teller.finance
+ */
+contract MarketsState is MarketsStateInterface, WhitelistedRole {
```
Juan-cc, line 20: When is WhiteListeRole set?

contracts/base/MarketsState.sol
```diff
@@ -0,0 +1,136 @@
...
+        @param borrowedAsset borrowed asset address.
+        @param collateralAsset collateral asset address.
+        @param amount amount to add.
+     */
+    function increaseRepayment(
+        address borrowedAsset,
+        address collateralAsset,
+        uint256 amount
+    ) external onlyWhitelisted() {
+        markets[borrowedAsset][collateralAsset].increaseRepayment(amount);
```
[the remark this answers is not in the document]
salazarguille: In all these cases, the events are emitted in the main function. Loans.takeOutLoan/createLoanWithTerms.

contracts/base/MarketsState.sol
```diff
@@ -0,0 +1,136 @@
...
+        @param borrowedAsset borrowed asset address.
+        @param collateralAsset collateral asset address.
+        @param amount amount to add.
+     */
+    function increaseSupply(
+        address borrowedAsset,
+        address collateralAsset,
+        uint256 amount
+    ) external onlyWhitelisted() {
+        markets[borrowedAsset][collateralAsset].increaseSupply(amount);
```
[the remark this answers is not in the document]
salazarguille: I updated the description. This is an initial version. We still need to add the invocations in the contract functions (like Loans.takeOutLoan, Lenders.deposit,  and others). We will not able to call the increase/decrease functions.

You are right. I need to add the decrease functions.

contracts/base/MarketsState.sol
salazarguille: This will be set in the deployment process.

test/utils/consts.js
```diff
@@ -89,4 +89,12 @@ module.exports = {
     printSeparatorLine: (length = 100, separator = '-') => {
         console.log(`\n${separator.repeat(length)}\n`);
     },
+    createMocks: async (Mock, total) => {
+        const result = [];
+        for (const index of _.range(0, total)) {
+            const mock = await Mock.new();
+            result.push(mock.address);
```
passabilities, line 96: Might be better to just return the mock instances rather than just the addresses.

test/base/MarketsStateGetMarketTest.js
salazarguille: Fixed

PR #113, opened 2020-08-08
[PROTOCOL-112] ATMGovernance contract

Juan-cc:
It includes:

- ATMGovernance contract
- IATMGovernance interface
- New unit tests
- Deploy script changed to include ATMGovernance deploy.

image


contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,163 @@
...
+    /* External Functions */
+
+    /**
+        @notice Adds a new General Setting to this ATM.
+        @param settingName name of the setting to be added.
+        @param settingValue value of the setting to be added.
+     */
+    function addGeneralSetting(bytes32 settingName, uint256 settingValue)
+        external
+        onlySigner
```
salazarguille, line 54: Please, use modifiers as functions like:
```
onlySigner()
whenNotPaused()
```

Similiar in other functions.

contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,163 @@
...
+        @param newValue new value to be set for this settingName. 
+     */
+    function updateGeneralSetting(bytes32 settingName, uint256 newValue)
+        external
+        onlySigner
+        whenNotPaused 
+    {
+        require(newValue > 0, "GENERAL_SETTING_MUST_BE_POSITIVE");
+        require(settingName != "", "GENERAL_SETTING_MUST_BE_PROVIDED");
+        uint256 oldVersion = generalSettings[settingName];
```
salazarguille, line 76: Should it be oldValue?

contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,163 @@
...
+import "@openzeppelin/contracts/access/roles/SignerRole.sol";
+
+// Interfaces
+import "./IATMGovernance.sol";
+
+/**
+    @notice This contract is used to modify Risk Settings, CRA or DataProviders for a specific ATM.
+    @author develop@teller.finance
+ */
+contract ATMGovernance is Pausable, SignerRole, IATMGovernance {
```
salazarguille, line 21: How do we deal with the pause feature? Who will able to call the pause function?
Do we need to inject the ATMSettings (it has the pause function) to verify whether the ATM is paused or not?

contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,163 @@
...
+        @param settingName name of the setting to be modified.
+        @param newValue new value to be set for this settingName. 
+     */
+    function updateGeneralSetting(bytes32 settingName, uint256 newValue)
+        external
+        onlySigner
+        whenNotPaused 
+    {
+        require(newValue > 0, "GENERAL_SETTING_MUST_BE_POSITIVE");
+        require(settingName != "", "GENERAL_SETTING_MUST_BE_PROVIDED");
```
salazarguille, line 75: Do you need to check if the new value != current one?

contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,163 @@
...
+
+    // List of general ATM settings. We don't accept settings equal to zero.
+    // Example: supplyToDebtRatio  => 5044 = percentage 50.44
+    // Example: supplyToDebtRatio => 1 = percentage 00.01
+    mapping(bytes32 => uint256) public generalSettings;
+
+
+    // List of Market specific Asset settings on this ATM
+    // Asset address => Asset setting name => Asset setting value
+    mapping(address => mapping(bytes32 => uint256)) public assetMarketSettings;
```
salazarguille, line 36: What would be some examples of this mapping?

contracts/ATM/IATMGovernance.sol
```diff
@@ -0,0 +1,240 @@
...
+        @param settingName name of the setting to be added.
+        @param settingValue value of the setting to be added.
+     */
+    function addAssetMarketSetting(address asset, bytes32 settingName, uint256 settingValue) external;
+
+    // /**
+    //     @notice Updates an existing Asset Setting from a specific Market on this ATM.
+    //     @param asset market specific asset address.
+    //     @param settingName name of the setting to be added.
+    //     @param settingValue value of the setting to be added.
```
salazarguille, line 173: Please, remove the commented code.

contracts/ATM/ATMGovernance.sol
salazarguille: As I saw in the unit tests, here is where we can configure the supplyToDebtRatio per market.
My question is how we define the market. It is a pair (borrowed/lent token and collateral token. Ex: DAI/ETH or USDC/ETH). But we only have one asset in this mapping.

test/ATM/ATMGovernanceRemoveGeneralSettingTest copy.js
```diff
@@ -0,0 +1,73 @@
+// JS Libraries
```
salazarguille, line 1: I think the filename is not ok.

contracts/ATM/ATMGovernance.sol
Juan-cc: Done!

contracts/ATM/ATMGovernance.sol
Juan-cc: corrected

contracts/ATM/ATMGovernance.sol
Juan-cc: Yes, based on yesterday's conversation I removed the Pausable capability and I'll inject ATMSettings

contracts/ATM/ATMGovernance.sol
Juan-cc: Yes, validation added

contracts/ATM/IATMGovernance.sol
Juan-cc: Oh that's to avoid compilation errors but I wanted to share with you the interface as it will be. I am removing comments as I am implementing methods.

test/ATM/ATMGovernanceRemoveGeneralSettingTest copy.js
Juan-cc: Filename modified

contracts/ATM/ATMGovernance.sol
Juan-cc: ATMGovernance is currently a contract where we store a specific ATM settings. I was following the rule to leave configurations/settings non-pausable in case we need to modify a value to fix an issue, so I removed Pausable capability and wasn't planning to add Settings. We might need a discussion around this.

contracts/ATM/ATMGovernance.sol
Juan-cc: As of today supplyToDebtRadio is one per ATM so it's configured as a General Setting (generalSettings['supplyToDebtRadio']).

Then, the assets we configure per market are assetMarketSettings[asset address][setting name] = value and I'll add some examples in the comments also.

contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,281 @@
...
+
+    /**
+        @notice Updates an existing Data Provider on a specific Data Type array.
+        @param dataTypeIndex array index for this Data Type.
+        @param oldProviderIndex previous data provider index.
+        @param newProvider new data provider address.
+     */
+    function updateDataProvider(
+        uint8 dataTypeIndex,
+        uint256 oldProviderIndex,
```
salazarguille, line 186: I think the var name shouldn't be "old". Maybe providerIndex?

contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,281 @@
...
+        address newProvider
+    ) external onlySigner() {
+        require(newProvider.isContract(), "DATA_PROVIDER_MUST_BE_A_CONTRACT");
+        address oldProvider = dataProviders[dataTypeIndex][oldProviderIndex];
```
salazarguille, line 190: should you validate if dataProviders[dataTypeIndex].length > providerIndex?

contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,281 @@
...
+        @notice Returns a Data Provider on a specific Data Type array.
+        @param dataTypeIndex array index for this Data Type.
+        @param dataProviderIndex data provider index number.
+     */
+    function getDataProvider(uint8 dataTypeIndex, uint256 dataProviderIndex)
+        external
+        view
+        returns (address)
+    {
+        require(
```
salazarguille, line 267: I think we don't need to validate it here (view function). In case the index is out of range it should return 0x0.

salazarguille:
Great job! I added some questions/comments.

contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,281 @@
...
+// Interfaces
+import "./IATMGovernance.sol";
+
+
+/**
+    @notice This contract is used to modify Risk Settings, CRA or DataProviders for a specific ATM.
+    @author develop@teller.finance
+ */
+contract ATMGovernance is SignerRole, IATMGovernance {
+    // TODO: Inject Settings
```
passabilities, line 22: need still?

contracts/ATM/ATMGovernance.sol
```diff
@@ -0,0 +1,281 @@
...
+        @notice Adds a new General Setting to this ATM.
+        @param settingName name of the setting to be added.
+        @param settingValue value of the setting to be added.
+     */
+    function addGeneralSetting(bytes32 settingName, uint256 settingValue)
+        external
+        onlySigner()
+    {
+        require(settingValue > 0, "GENERAL_SETTING_MUST_BE_POSITIVE");
+        require(settingName != "", "GENERAL_SETTING_MUST_BE_PROVIDED");
```
passabilities, line 61: Looks like this is repeated a bit, maybe a modifier?

migrations/2_deploy.js
```diff
@@ -46,9 +49,15 @@ module.exports = async function(deployer, network, accounts) {
   const currentBlockNumber = await web3.eth.getBlockNumber();
 
   await deployerApp.deploys([ZDAI, ZUSDC], txConfig);
-  console.log(`Deployed tokens: ZDAI [${ZDAI.address}] ZUSDC [${ZUSDC.address}] `);  
-  
+
+  console.log(`Deployed tokens: ZDAI [${ZDAI.address}] ZUSDC [${ZUSDC.address}] `);
+
+  // ATM Deployments
+  await deployerApp.deploy(ATMGovernance, txConfig); // TODO: add Gnosis multisig as signer/pauser (not the DAO for now)
```
passabilities, line 56: Take a look at the `feature/upgradable` branch for how to deploy an upgradable version of this. I can help if needed.

contracts/ATM/ATMGovernance.sol
Juan-cc: we will not pause this contract because it mainly contains ATM Settings so I'll remove this comment.

contracts/ATM/ATMGovernance.sol
Juan-cc: I really like this improvement, I created a task to implement this suggestion: https://teller-finance.atlassian.net/browse/PROTOCOL-172

migrations/2_deploy.js
Juan-cc: This will be replaced by the ATMFactory

PR #115, opened 2020-08-11
[PROTOCOL-135, PROTOCOL-141] Create ATM Settings Contract

salazarguille:
It includes:

Note: As I needed the ATMGovernanceFactoryInterface, I copied/pasted from another PR.

- Ability to pause an ATM.
- Configuration in the ATM settings contract.
- New unit tests.

image

### DIFF
--- a/contracts/settings/ATMSettings.sol
+++ b/contracts/settings/ATMSettings.sol
@@ -11,6 +11,7 @@ import "../interfaces/SettingsInterface.sol";
 import "../atm/ATMGovernanceFactoryInterface.sol";
 import "./ATMSettingsInterface.sol";
 
+
 /**
     @notice It manages the settings for the ATMs.
 
@@ -18,7 +19,7 @@ import "./ATMSettingsInterface.sol";
  */
 contract ATMSettings is ATMSettingsInterface {
     using Address for address;
-    
+
     /** Constants */
 
     /* State Variables */
@@ -66,7 +67,10 @@ contract ATMSettings is ATMSettingsInterface {
     /* Constructor */
 
     constructor(address atmGovernanceFactoryAddress, address settingsAddress) public {
-        require(atmGovernanceFactoryAddress != address(0x0), "ATM_GOV_FACTORY_MUST_BE_PROVIDED");
+        require(
+            atmGovernanceFactoryAddress != address(0x0),
+            "ATM_GOV_FACTORY_MUST_BE_PROVIDED"
+        );
         require(settingsAddress != address(0x0), "SETTINGS_MUST_BE_PROVIDED");
 
         atmGovernanceFactory = ATMGovernanceFactoryInterface(atmGovernanceFactoryAddress);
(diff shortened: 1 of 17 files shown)


contracts/settings/ATMSettings.sol
```diff
@@ -0,0 +1,243 @@
...
+    SettingsInterface public settings;
+
+    ATMGovernanceFactoryInterface public atmGovernanceFactory;
+
+    /**
+        @notice It represents a mapping to identify whether a ATM is paused or not.
+
+        i.e.: address(ATM) => true or false.
+     */
+    mapping(address => bool) public atmToPaused;
```
passabilities, line 36: The naming of this mapping is confusing. Maybe `atmPaused`?

contracts/settings/ATMSettings.sol
```diff
@@ -0,0 +1,243 @@
...
+    /**
+        @notice It pauses an given ATM.
+        @param atmAddress ATM address to pause.
+     */
+    function pauseATM(address atmAddress)
+        external
+        withPauserRole()
+        withValidATM(atmAddress)
+    {
+        require(settings.isPaused() == false, "PLATFORM_IS_ALREADY_PAUSED");
```
passabilities, line 91: Should we allow ATMs to be un/paused separately from the platform? That way if they need to still be paused even if we unpause the platform

contracts/settings/ATMSettings.sol
```diff
@@ -0,0 +1,243 @@
...
+        emit MarketToAtmSet(borrowedToken, collateralToken, atmAddress, msg.sender);
+    }
+
+    /**
+        @notice Updates a new ATM for a given market (borrowed token and collateral token).
+        @param borrowedToken borrowed token address.
+        @param collateralToken collateral token address.
+        @param newAtmAddress the new ATM address to update.
+     */
+    function updateATMToMarket(
```
passabilities, line 154: Given we are going to make the ATMs upgradeable, do we want to be able to change the ATM proxy address still?

contracts/settings/ATMSettings.sol
salazarguille: Yes, it is possible. We don't want to pause an atm when the platform is paused or unpause an atm when the platform is paused. Even, when you call the isATMPaused function it verifies both (platform or atm).

contracts/settings/ATMSettings.sol
salazarguille: In this case yes, because we might use a different ATM (insurance atm instead of real state) for a given market.

test/settings/ATMSettingsConstructorTest.js
```diff
@@ -0,0 +1,43 @@
...
+    let mocks;
+    
+    beforeEach('Setup for each test', async () => {
+        mocks = await createMocks(Mock, 10);
+    });
+
+    withData({
+        _1_basic: [1, 4, undefined, false],
+        _2_empty_atm_factory: [-1, 5, 'ATM_GOV_FACTORY_MUST_BE_PROVIDED', true],
+        _3_empty_settins: [3, -1, 'SETTINGS_MUST_BE_PROVIDED', true],
```
Juan-cc, line 21: Typo: _3_empty_settins => _3_empty_settings

contracts/settings/ATMSettings.sol
```diff
@@ -0,0 +1,243 @@
...
+        @notice It pauses an given ATM.
+        @param atmAddress ATM address to pause.
+     */
+    function pauseATM(address atmAddress)
+        external
+        withPauserRole()
+        withValidATM(atmAddress)
+    {
+        require(settings.isPaused() == false, "PLATFORM_IS_ALREADY_PAUSED");
+        require(atmToPaused[atmAddress] == false, "ATM_IS_ALREADY_PAUSED");
```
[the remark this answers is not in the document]
salazarguille: Correct.

- You only can pause an ATM when the platform is not paused (the isATMPaused has both conditions): we don't want to an unpaused ATM when the platform is paused (due to an issue).
- You only can unpause an ATM when the platform is not paused (the isATMPaused has both conditions): similar here.
 
Any third contract that verifies whether an atm is paused, will verify both conditions is platform paused or is atm paused internally.

contracts/settings/ATMSettings.sol
```diff
@@ -0,0 +1,243 @@
...
+        emit ATMUnpaused(atmAddress, msg.sender);
+    }
+
+    /**
+        @notice Gets whether an ATM is paused (or the platform is paused) or not.
+        @param atmAddress ATM address to test.
+        @return true if ATM is paused. Otherwise it returns false.
+     */
+    function isATMPaused(address atmAddress) external view returns (bool) {
+        return settings.isPaused() || atmToPaused[atmAddress];
```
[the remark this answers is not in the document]
salazarguille: We wouldn't able to pause an ATM if the platform is paused.

PR #116, opened 2020-08-11
[PROTOCOL-119] Adding Interest Validator

salazarguille:
It includes:

- A new InterestValidator interface to be used in the LendingPool.withdrawInterest function. By default, it will be empty. So we will not be validating the amount to withdraw. But it allows us to implement a new version in the future.
- New unit tests

image

### DIFF
--- a/contracts/base/Lenders.sol
+++ b/contracts/base/Lenders.sol
@@ -64,32 +64,6 @@ contract Lenders is Base, LendersInterface {
     /** External Functions */
 
     /**
-        @notice It initializes this contract instance.
-        @param zTokenAddress zToken contract address.
-        @param lendingPoolAddress lending pool contract address.
-        @param interestConsensusAddress interest consensus contract address.
-        @param settingAddress settings contract address.
-        @param marketsAddress markets contract address.
-     */
-    function initialize(
-        address zTokenAddress,
-        address lendingPoolAddress,
-        address interestConsensusAddress,
-        address settingAddress,
-        address marketsAddress
-    ) external isNotInitialized() {
-        zTokenAddress.requireNotEmpty("ZTOKEN_MUST_BE_PROVIDED");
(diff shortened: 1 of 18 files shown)


contracts/base/Lenders.sol
```diff
@@ -155,6 +129,32 @@ contract Lenders is Base, LendersInterface {
...
+        lendingPoolAddress.requireNotEmpty("LENDING_POOL_MUST_BE_PROVIDED");
+        interestConsensusAddress.requireNotEmpty("CONSENSUS_MUST_BE_PROVIDED");
+
+        _initialize(settingAddress, marketsAddress);
+
+        zToken = zTokenAddress;
+        lendingPool = lendingPoolAddress;
+        interestConsensus = InterestConsensusInterface(interestConsensusAddress);
+    }
+
```
passabilities, line 157: Any reason this was moved down here? Makes more sense to keep it towards the top.

contracts/base/LendingPool.sol
```diff
@@ -238,8 +216,76 @@ contract LendingPool is Base, LendingPoolInterface {
...
+        _initialize(settingsAddress, marketsAddress);
+
+        zToken = ZTokenInterface(zTokenAddress);
+        lendingToken = IERC20(lendingTokenAddress);
+        lenders = LendersInterface(lendersAddress);
+        loans = loansAddress;
+        cToken = cTokenAddress;
+        interestValidator = InterestValidatorInterface(interestValidatorAddress);
+    }
+
```
passabilities, line 282: Again should leave the `initializer` towards the top.

contracts/base/Lenders.sol
salazarguille: Yes, this is due to gas consumption reasons. The functions that will be invoked more frequently should be on the top. So they consume less gas when EVM compares the sig function.

contracts/base/LendingPool.sol
salazarguille: The same reason.

PR #118, opened 2020-08-13
[PROTOCOL-30] Make contracts upgradeable.

passabilities:
It is:
- Uses proxies to make contracts upgrdeable
- Tests that contracts can be upgraded.

image


contracts/mock/base/InitializableModifiersMock.sol
```diff
@@ -1,12 +1,12 @@
 pragma solidity 0.5.17;
```
salazarguille, line 1: Please rename filename to TInitializableModifiersMock.sol

test/UpgradeableProxyTest.js
```diff
@@ -0,0 +1,65 @@
...
+
+        v1InitData = v1LibraryInstance.contract.methods.initialize(initValue).encodeABI();
+    });
+
+    withData({
+        _1_only_admin_upgrade: [ accounts[1], accounts[0], false, true, "UPGRADABLE_CALLER_MUST_BE_ADMIN" ],
+        _2_successful: [ accounts[0], accounts[0], false, false, null ],
+        _3_initialize_after_constructor: [ accounts[0], accounts[0], true, false, null ]
+    }, function(caller, admin, initAfter, mustFail, expectedErrorMessage) {
+        it(t("user", "transfer", "Should be able transfer tokens.", mustFail), async function() {
```
salazarguille, line 30: The function name and description don't match with  the unit test.

test/UpgradeableProxyTest.js
```diff
@@ -0,0 +1,65 @@
...
+                }
+
+                const implementationV1 = await proxy.implementation.call();
+                assert.equal(implementationV1.toLowerCase(), v1LibraryInstance.address.toLowerCase(), "V1 implementation addresses do not match");
+
+                const v1Value = await v1.value.call();
+                assert.equal(initValue.toString(), v1Value.toString(), "Initial values do not match");
+
+                await proxy.upgradeTo(v2LibraryInstance.address, { from: caller });
+                const implementationV2 = await proxy.implementation.call();
```
salazarguille, line 50: So, as I see, we  always need to get the current implementation using the proxy.implementation() function. Right? So, then, if we update it (using proxy.upgradeTo(...)), we get the new implementation.

test/UpgradeableProxyTest.js
```diff
@@ -0,0 +1,65 @@
...
+                assert.equal(implementationV2.toLowerCase(), v2LibraryInstance.address.toLowerCase(), "V2 implementation addresses do not match");
+
+                const newValue = 1234;
+                await v2.setValue(newValue);
+
+                const updatedValue = await v1.value.call();
```
salazarguille, line 56: What would happen if I update the value using the proxy? 
```proxy.setValue(newValue)``` Should it update the value in the current implementation?

L56/L57 why did the proxy update the value in the V1? if you already had upgraded the implementation (to V2). Can we assert the value for the V2? I think it should have the new value.

File: test/UpgradeableProxyTest.js [hunk 5]
```diff
@@ -0,0 +1,65 @@
...
+
+                const v1Value = await v1.value.call();
+                assert.equal(initValue.toString(), v1Value.toString(), "Initial values do not match");
+
+                await proxy.upgradeTo(v2LibraryInstance.address, { from: caller });
+                const implementationV2 = await proxy.implementation.call();
+                assert.equal(implementationV2.toLowerCase(), v2LibraryInstance.address.toLowerCase(), "V2 implementation addresses do not match");
+
+                const newValue = 1234;
+                await v2.setValue(newValue);
```
salazarguille, line 54: Regarding the ether/erc20 balances. How this proxy deal with it? I think we should have unit tests to verify it.

test/utils/settings-helper.js
```diff
@@ -21,7 +23,11 @@ const createSettingsInstance = async (
     Settings,
     params,
 ) => {
-    const instance = await Settings.new();
+    const { from } = Settings.class_defaults
```
salazarguille, line 26: What is class_defaults?

salazarguille:
Please, resolve the compilation error. See details here: https://github.com/teller-protocol/teller-protocol-v1/runs/978598903?check_suite_focus=true#step:5:17

test/UpgradeableProxyTest.js
passabilities: The `implementation` function tells us which version of the contract we are currently using by pointing to the contract that holds all the functionality. However, you don't normally need to know what the current implementation is to interact with the contract. The address that is given via the `proxy (AdminUpgradeabilityProxy)` contract is the main entry point that forwards any calls to the current `implementation`.

test/UpgradeableProxyTest.js
passabilities: So `v1` and `v2` point to the same (proxy) contract but with different ABIs. They both have a `value` property but only `v2` has the `setValue` function on it. The `proxy` object only contains the information `implementation` which points to either the `v1` or `v2` contract and the function `upgradeTo` that allows us to change the currently contract implementation.

test/UpgradeableProxyTest.js
passabilities: I'll create a unit test for that. But they won't change. because the balances will be on the `proxy` address which is the only address that we need to know in order to interact with the contract. It just forwards any call to the current `implementation` contract to run the current version of functionality.

test/utils/settings-helper.js
passabilities: Its an object that is set by Truffle that holds which address should be the `from` and the `gasLimit` of the transaction when when you do `await Settings.new()`. Since we don't have access to the `accounts` object like in the test files, this was the easiest way to grab a valid account address.

passabilities:
Will add test coverage once complete :sweat_smile:

contracts/base/EtherCollateralLoans.sol
```diff
@@ -114,6 +112,8 @@ contract EtherCollateralLoans is LoansBase {
             settingsAddress,
             marketsAddress
         );
+
+        collateralToken = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
```
Juan-cc, line 116: Did we remove type and scope?

contracts/base/UpgradeableProxy.sol
```diff
@@ -0,0 +1,14 @@
+import "@openzeppelin/upgrades/contracts/upgradeability/AdminUpgradeabilityProxy.sol";
+
+contract UpgradeableProxy is AdminUpgradeabilityProxy {
+    modifier ifAdmin() {
```
Juan-cc, line 4: Probably we could use ifAdmin() from BaseAdminUpgradeabilityProxy unless we need to throw 'UPGRADABLE_CALLER_MUST_BE_ADMIN' as a message.

migrations/2_deploy.js
```diff
@@ -4,6 +4,7 @@ const PoolDeployer = require('./utils/PoolDeployer');
 const initSettings = require('./utils/init_settings');
 
 const ERC20 = artifacts.require("@openzeppelin/contracts/token/ERC20/ERC20Detailed.sol");
+const AdminUpgradeabilityProxy = artifacts.require("./base/UpgradeableProxy.sol");
```
Juan-cc, line 7: We could log a message including AdminUpgradeabilityProxy's address

migrations/3_deploy_mocks.js
```diff
@@ -1,5 +1,6 @@
 const assert = require('assert');
 const DeployerApp = require('./utils/DeployerApp');
+const AdminUpgradeabilityProxy = artifacts.require("./base/UpgradeableProxy.sol");
```
Juan-cc, line 3: We could add a message logging Admin's address

package.json
```diff
@@ -52,6 +52,7 @@
   "dependencies": {
     "@chainlink/contracts": "^0.0.5",
     "@openzeppelin/contracts": "2.5.1",
+    "@openzeppelin/contracts-ethereum-package": "^2",
```
Juan-cc, line 55: should we use a strict version number?

test/UpgradeableProxyTest.js
Juan-cc: It would be great if we could have a test to validate we can freely access/transfer eth from both versions V1 and V2. Let's say accumulate eth using V1 (should remain in proxy), then upgrade to V2 and transfer some of those funds.

contracts/base/EtherCollateralLoans.sol
passabilities: It is defined in `LoansBase.sol`. Contracts that are **upgradable** should not be instantiated outside the `initialize` function except for **constant**s.

test/UpgradeableProxyTest.js
passabilities: I can definitely create a test to demonstrate this but the funds will only be interacting with the `proxy.address`. Notice these lines: https://github.com/teller-protocol/teller-protocol-v1/pull/118/files/6ce23521c0e7a4ab608687abc9d5fd504ff0ca74#diff-b66228bea33641cbcf8a0c2f1387bcb7R35-R36

migrations/2_deploy.js
passabilities: We won't be deploying this address on its own. It will only be deployed when when want to create a brand new upgradable contract. Also for the ones we are creating they are being logged via the DeployerApp [here](https://github.com/teller-protocol/teller-protocol-v1/pull/118/files/6ce23521c0e7a4ab608687abc9d5fd504ff0ca74#diff-0c3e3d003115713a42f37b1f3e04c46fR18-R19)

migrations/3_deploy_mocks.js
passabilities: You mean which account is the `admin` for the upgradeable proxies?

package.json
passabilities: Yea sure I'll update. I believe **3** is already out.

contracts/base/UpgradeableProxy.sol
passabilities: We want our own to have the `require` statement instead of `if` so that if an admin is required for a function it will revert rather than call the fallback. Which is the intended functionality we are looking for here.

migrations/3_deploy_mocks.js
Juan-cc: I thought we would have one ProxyAdmin contract to reuse for all the contracts. As it's one per upgradeable contract I don't think we need to log every admin for every upgradeable contract.

contracts/mock/upgradable/UpgradableV2.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity 0.5.17;
+
+import "./UpgradableV1.sol";
+
+
+contract UpgradableV2 is UpgradableV1 {
+    function setValue(uint256 newValue) external {
```
Juan-cc, line 7: Right now we are validating we can change previous storage, one other case we could test is to extend storage allocation? For example V1 has 1 state variable, V2 adds 1 state variable, V3 adds 1 state variable, and we validate that V3 can change any V1, V2 and V3 variables.

package.json
```diff
@@ -52,6 +52,8 @@
   "dependencies": {
     "@chainlink/contracts": "^0.0.5",
     "@openzeppelin/contracts": "2.5.1",
+    "@openzeppelin/contracts-ethereum-package": "2.5.0",
+    "@openzeppelin/upgrades": "^2.8.0",
```
Juan-cc, line 56: Should we use strict version number here? 2.8.0

test/base/UpgradeableProxyUpgradeToTest.js
```diff
@@ -0,0 +1,128 @@
...
+
+    before(async () => {
+        dai = await DAI.new()
+    })
+
+    beforeEach("Setup for each test", async () => {
+        v1LibraryInstance = await UpgradableV1.new();
+        v2LibraryInstance = await UpgradableV2.new();
+
+        v1InitData = v1LibraryInstance.contract.methods.initialize(initValue).encodeABI();
```
Juan-cc, line 27: I see we need to encode the ABI and we have several contracts using ABI encoder v2, do we know if there will be no problem using V2 enconder?

package.json
passabilities: I believe they are on 3 now for solc 6 but yea couldn't hurt to make it strict.

test/base/UpgradeableProxyUpgradeToTest.js
passabilities: It is only necessary to encode the ABI of the `initialize`er functions on upgradeable contracts that we want to deploy and then immediately call the initialize function on them. Otherwise if you don't want to deploy the contract and call a function on it at the same time, you can just pass 0 bytes (`0x`) as the 3rd argument to the upgradeable proxy as you can see in this test a few lines down on line 38-39:
```
                const initData = initAfter ? "0x" : v1InitData;
                const proxy = await AdminUpgradeabilityProxy.new(v1LibraryInstance.address, admin, initData);
```

contracts/mock/upgradable/UpgradableV2.sol
passabilities: Should we add these test in your branch with the upgradeable ATM contracts?

Juan-cc: At the moment we don't need to change ATMGovernance state so I wouldn't add that in that branch. I think it'll be simpler to include this test on current branch because contracts are simpler (UpgradableV2, UpgradableV1)

test/base/UpgradeableProxyUpgradeToTest.js
Juan-cc: Yes, my question is about making sure we don't end up having encoding issues on the contracts that use ABIEncoderV2. For example InterestConsensus uses ABIEncoderV2 and we will make it upgradeable. In the case we want to call initialize() during deployment (which makes sense) we will call initialize(params).encodeABI(). Is there something we can do to validate we won't have encoding issues on ABIEncoderV2 contracts?

PR #122, opened 2020-08-14
Fixes: ATM Token Compilation Issues

salazarguille:
It fixes some compilation warnings in the ATM Token contract.


contracts/atm/ATMToken.sol
```diff
@@ -178,7 +178,7 @@ contract ATMToken is ATMTokenInterface, ERC20Detailed, ERC20Mintable, ERC20Burna
      *
      * - minted tokens must not cause the total supply to go over the cap.
      */
-    function _beforeTokenTransfer(address from, address to, uint256 amount)
+    function _beforeTokenTransfer(address from, address, uint256 amount)
         internal
         view
         returns (bool)
```
passabilities, line 184: Looks like this function is supposed to return a bool?

PR #138, opened 2020-08-24
[PROTOCOL-193] [PROTOCOL-194] Compound + Uniswap modifications and unit tests

Juan-cc:
It includes:
 - Uniswap Dapp unit tests.
 - Uniswap comments added.
 - Compound Dapp unit tests.
 - Compound comments added.

image


package-lock.json
```diff
@@ -0,0 +1,14186 @@
+{
```
passabilities, line 1: Since we are just using `yarn` lets delete `npm`'s lock file.

contracts/base/Escrow/Dapps/Uniswap.sol
```diff
@@ -18,75 +24,97 @@ import "./IDApp.sol";
...
+        @param path An array of token addresses. path.length must be >= 2. Pools for each consecutive pair of addresses must exist and have liquidity.
+        @param sourceAmount amount of source element (ETH or Tokens) to swap.
+        @param minDestination The minimum amount of output tokens that must be received for the transaction not to revert.
+        @dev This function mainly invokes 3 Uniswap external functions:
+            https://uniswap.org/docs/v2/smart-contracts/router02/#swapexactethfortokens
+            https://uniswap.org/docs/v2/smart-contracts/router02/#swapexacttokensforeth
+            https://uniswap.org/docs/v2/smart-contracts/router02/#swapexacttokensfortokens
+     */
+    function swap(
+        address canonicalWeth,
```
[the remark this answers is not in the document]
Juan-cc: Initially we had this as a constant but we need to change it on each network, so as it can't be a state variable in the dApp (delegatecall) we decided to pass it as a function parameter. Please feel free to suggest a different approach

contracts/base/Escrow/Dapps/Uniswap.sol
```diff
@@ -18,75 +24,97 @@ import "./IDApp.sol";
...
-            require(address(this).balance >= sourceAmount, "UNISWAP_INSUFFICIENT_SOURCE");
-
-            router.swapExactETHForTokens(0, path, address(this), now);
+        require(minDestination > 0, "UNISWAP_MIN_DESTINATION_ZERO"); // what if there is no minimum?
+
+        uint256[] memory amounts;
+        uint256 balanceBeforeSwap = 0;
+        uint256 balanceAfterSwap = 0;
+        balanceBeforeSwap = IERC20(destination).balanceOf(address(this));
+        if (source == canonicalWeth) {
```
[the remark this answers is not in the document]
Juan-cc: Yes that would be one option, I think we should implement something like that but I think we should create a new task to implement that

package.json
```diff
@@ -39,11 +43,11 @@
...
     "husky": "^4.2.3",
     "jsonfile": "^6.0.1",
-    "leche": "^2.3.0",
     "lodash": "^4.17.19",
-    "mocha": "^7.1.0",
+    "mocha": "6.2.2",
+    "leche": "^2.3.0",
     "openzeppelin-test-helpers": "^0.5.1",
-    "prettier": "^2.0.2",
+    "prettier": "1.19.1",
```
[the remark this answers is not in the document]
Juan-cc: Downgrade was made due to incompatible dependencies, please if you are not getting any dependency error and are using a 2.0.2 version feature let me know to revert this change

package-lock.json
Juan-cc: Yes, my bad committing that I'll fix it

package.json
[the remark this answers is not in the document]
Juan-cc: Changes were finally reverted to previous versions

contracts/base/Escrow/Dapps/Uniswap.sol
```diff
@@ -18,75 +24,99 @@ import "./IDApp.sol";
...
+    @author develop@teller.finance
+ */
+contract Uniswap is IUniswap {
     using AddressLib for address;
+    using Address for address;
+
+    /* State Variables */
+    // State is shared with Escrow contract as it uses delegateCall() to interact with this contract.
+    
+     function() external payable {}
```
passabilities, line 39: Dapp contracts should not have a fall back function. More importantly, they should not be able to hold ANY funds.

contracts/base/Escrow/Dapps/Uniswap.sol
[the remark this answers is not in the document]
passabilities: Created a task in the backlog :+1:

test/base/Escrow/Dapps/UniswapSwapTest.js
```diff
@@ -1,59 +1,158 @@
...
+    routerIsContract,
+    wethIsContract,
+    mustFail,
+    expectedErrorMessage
+  ) {
+    it(t("uniswap", "swapExactETHForTokens", "Should be able (or not) to swap tokens on Uniswap", mustFail), async function() {
+      // Setup
+      if (sourceBalance > 0) {
+        if (path[0] === 'eth') {
+          await web3.eth.sendTransaction({ to: instance.address, from: accounts[0], value: sourceBalance + 1 })
```
passabilities, line 58: As mentioned above, we don't the dapp contracts to accept any funds. For this test it makes sense, however, instead of having a fallback function on the dapp contract itself, we should create a mock contract that extends it with just a payable fallback function to accept ETH so we don't deploy the main dapp with one.

contracts/base/Escrow/Dapps/Uniswap.sol
Juan-cc: Payable fallback function added on Mock for tests only.

test/base/Escrow/Dapps/UniswapSwapTest.js
Juan-cc: Same as above, payable fallback moved to Mock only.

PR #148, opened 2020-08-29
[PROTOCOL-232]: Improving Contract Versions and Markets Creation

salazarguille:
Initial version


contracts/mock/base/LoansBaseMock.sol
```diff
@@ -71,16 +71,13 @@ contract LoansBaseMock is LoansBase {
         address lendingPoolAddress,
         address loanTermsConsensusAddress,
         address settingsAddress,
-        address marketsAddress,
-        address atmSettingsAddress
+        address
```
sypeer, line 74: Why is this empty parameter here?

PR #154, opened 2020-09-02
Test fixes 4

sypeer:


test/base/UpgradeableProxyUpgradeToTest.js
```diff
@@ -1,139 +1,80 @@
...
     beforeEach("Setup for each test", async () => {
-        
-        v1LibraryInstance = await UpgradableV1.new();
-        await v1LibraryInstance.initialize(initValue);
-        v2LibraryInstance = await UpgradableV2.new();
-        await v2LibraryInstance.initialize(initValue);
-
-        v1InitData = v1LibraryInstance.contract.methods.initialize(initValue).encodeABI();
+        const owner = accounts[0];
+        settingsLogic = await createTestSettingsInstance(
```
passabilities, line 22: since this is being used here as the actual settings contract and not the logic, lets rename to just `settings`

sypeer: Renamed

PR #166, opened 2020-09-06
CI: test only changed or new files

tpscrpt:
Of note:
 - We can do quite expressive conditional testing at runtime using nodejs (fuzzy search under test/ for specific tokens)
 - Coverage is not accounted for in this first example

I don't think I would want this on my `develop`, but I might want it on a separate `fixes` branch for small fixes which need to be dispatched quickly or which you believe won't crash when being merged in `develop`. If it does crash when making a PR from `fixes` to `develop`, we can easily just `git blame` and put it on the scoreboard.

For now, it will just run as a second job on PRs to `develop`, taking around 3 to 5 minutes if number of changed test files is ~10.

Refs:
 1. Only changed test files being tested - https://github.com/tpscrpt/teller-protocol-v1/runs/1078791412?check_suite_focus=true
 2. Job fails if a test fails - https://github.com/tpscrpt/teller-protocol-v1/runs/1078802480?check_suite_focus=true
 3. State of a PR after latest commit is a failure - https://github.com/tpscrpt/teller-protocol-v1/pull/3
 4. On this PR, it looks like - https://github.com/teller-protocol/teller-protocol-v1/pull/166/checks?check_run_id=1078863116#step:6:15
image

### DIFF
--- a/.github/workflows/develop-pr.yml
+++ b/.github/workflows/develop-pr.yml
@@ -31,6 +31,8 @@ jobs:
         name: Get List of Changed Files
         uses: jitterbit/get-changed-files@v1
       - name: Test Changes or New Files
+        env:
+          CHANGED_FILES: "{{ steps.files.outputs.all }}"
         run: |
-          git remote add base git@github.com:teller-protocol/teller-protocol-v1
-          yarn test:coverage `{{ steps.files.outputs.all }} | grep 'test/' | tr '\n' " "`
+          echo $CHANGED_FILES; echo 'abc'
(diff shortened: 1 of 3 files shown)


sypeer (approved):

salazarguille (approved):

tpscrpt:
Update: Now doesn't run tests if no changes in the test/ folder
This is not necessarily what is planned in the future but a convenience for now

In the future it would be nice to organize changed files by tags and have certain classes of tests be triggered based on those values.

PR #176, opened 2020-09-10
Fix/ci: dont exit on changed files error

tpscrpt:
The Github action I'm using to get a list of changed files is not the best and spits an error (despite correctly identifying which files were changed) under some circumstances.

This fix just adds `continue-on-error: true` for this job.

### DIFF
--- a/.github/workflows/develop-pr.yml
+++ b/.github/workflows/develop-pr.yml
@@ -1,4 +1,4 @@
-name: Test Changed or New Files
+name: Test Changed or New Files under test/
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


salazarguille (approved):

tpscrpt:
@salazarguille I'm making a few fixes I'll just push them to this branch

tpscrpt:
Newst commit message:
```
Fix/ci: Multiple bugs with develop-pr implementation

 - continue-on-error added for PR's with old `develop` base
 - use added_modified instead of all so we don't test non-existant files
 - fix use of quotes in a few places in the yml file

Feat/ci: Only install if tests will run as well

This should fix some of the issues I've been seeing in the actions and
issues which were reported on slack
```

tpscrpt:
G2G again

PR #177, opened 2020-09-10
DRAFT - Feature/add liquidity mining

Juan-cc:
It includes:
 - Initial liquidity mining design using staking.


Juan-cc:
This was used to present design to the team. Closing as other branch will be use for implementation.

PR #181, opened 2020-09-18
[PROTOCOL-275] Add repay loan to Escrow.

passabilities:
It: 
* Adds loan repay function to Escrow.
* When loan closed with an Escrow, transfer Escrow tokens to borrower or liquidator.
* Removes functionality to purchase the escrow.


contracts/base/Escrow.sol
```diff
@@ -165,16 +157,37 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
...
-        require(canPurchase(), "ESCROW_INELIGIBLE_TO_PURCHASE");
+    function repay(uint256 amount) external onlyOwner {
+        IERC20 token = IERC20(loans.lendingToken());
+        uint256 balance = _balanceOf(loans.lendingToken());
+        if (amount > balance) {
+            token.transferFrom(msg.sender, address(this), amount.sub(balance));
+        }
+        token.approve(address(loans), amount);
+
+        loans.repay(amount, loanID);
```
Juan-cc, line 172: We are emitting an event (LoanRepaid) on LoanBase, we might need to emit on Escrow and LoansBase to know who called LoansBase.repay().

test/base/EscrowRepayTest.js
```diff
@@ -0,0 +1,74 @@
...
+
+      try {
+        // Invocation
+        await instance.repay(amount, { from: caller });
+
+        const loansRepayCallCount = await loans.invocationCountForMethod.call(loansEncoder.encodeRepay());
+
+        // Assertions
+        assert.equal(loansRepayCallCount.toString(), "1", "Loans.repay was not called.");
+        assert(!mustFail);
```
Juan-cc, line 67: If we decide to emit an event, check that event was emitted.

contracts/base/Escrow.sol
```diff
@@ -165,16 +157,37 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
...
+        @dev If the Escrow's balance of the borrowed token is less than the amount to repay, transfer tokens from the sender's wallet.
+        @dev Only the owner of the Escrow can call this. If someone else wants to make a payment, they should call the loans contract directly.
      */
-    function purchaseLoanDebt() external payable {
-        require(canPurchase(), "ESCROW_INELIGIBLE_TO_PURCHASE");
+    function repay(uint256 amount) external onlyOwner {
+        IERC20 token = IERC20(loans.lendingToken());
+        uint256 balance = _balanceOf(loans.lendingToken());
+        if (amount > balance) {
+            token.transferFrom(msg.sender, address(this), amount.sub(balance));
```
salazarguille, line 168: I think we should validate the return value from the transferFrom call.

contracts/base/Escrow.sol
```diff
@@ -165,16 +157,37 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
...
+        }
+        token.approve(address(loans), amount);
```
salazarguille, line 170: Should we validate the return value?

contracts/base/Escrow.sol
```diff
@@ -165,16 +157,37 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
...
+        @dev The loan must not be active.
+        @dev The recipient must either be the loan borrower OR the loan must be already liquidated.
+        @param recipient address to send the tokens to.
+    */
+    function claimTokens(address recipient) external {
+        require(getLoan().status != TellerCommon.LoanStatus.Active, "LOAN_ACTIVE");
+        require(recipient == getBorrower() || getLoan().liquidated, "LOAN_NOT_LIQUIDATED");
+
+        address[] memory tokens = getTokens();
+        for (uint256 i = 0; i < tokens.length; i++) {
```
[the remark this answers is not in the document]
salazarguille: I agree. We should have an EstimateGasEscrowClaimTokensTest.js file and a table to know how many tokens we can process. It already created a new task for this sprint.

test/base/EscrowRepayTest.js
salazarguille: Don't we need to validate any amount after the repayment?

salazarguille (changes requested):

contracts/base/Escrow.sol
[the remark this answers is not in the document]
salazarguille: Should we validate the ```_balanceOf(tokens[i]) > 0```?

salazarguille (changes requested):

contracts/base/Escrow.sol
```diff
@@ -165,16 +157,37 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
...
+        @notice Repay this Escrow's loan.
+        @dev If the Escrow's balance of the borrowed token is less than the amount to repay, transfer tokens from the sender's wallet.
+        @dev Only the owner of the Escrow can call this. If someone else wants to make a payment, they should call the loans contract directly.
      */
-    function purchaseLoanDebt() external payable {
-        require(canPurchase(), "ESCROW_INELIGIBLE_TO_PURCHASE");
+    function repay(uint256 amount) external onlyOwner {
+        IERC20 token = IERC20(loans.lendingToken());
+        uint256 balance = _balanceOf(loans.lendingToken());
+        if (amount > balance) {
```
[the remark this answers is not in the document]
passabilities: I don't think the price really matters for the lending token since the borrower will need to repay the full amount no matter what.

contracts/base/Escrow.sol
passabilities: There's really no why to validate that the response will be false. The `transferFrom` function only throws if there is a problem or returns `true` so I mean I can add it but seems like it will just be redundant.

contracts/base/Escrow.sol
passabilities: Same here. It either throws or returns `true`. No `false` verification.

contracts/base/Escrow.sol
passabilities: The only one that can call `Escrow.repay` is the owner (aka the borrower).

test/base/EscrowRepayTest.js
passabilities: If we did emit an event, what should the values be? Emitting an event here and on `LoansBase` seems redundant to me. Only thing I can think that would be useful would be the Escrow's balance of the token? Everything else is already emitting from the LoansBase contract.

passabilities: > If we decide to emit an event, check that event was emitted.

If we did emit an event, what should the values be? Emitting an event here and on LoansBase seems redundant to me. Only thing I can think that would be useful would be the Escrow's balance of the token? Everything else is already emitting from the LoansBase contract.

> Don't we need to validate any amount after the repayment?

There's not much I can do in regards to validating amounts since all I am doing in the `Escrow.repay` function is:
1. Checking the Escrow's and borrower's balance (if the Escrow does not have enough funds, it trys to grab funds from the borrower)
1. Setting an allowance on the Loans contract for the repayment amount
1. Calling `Loans.repay`

So unless I mock the `Loans.repay` to pull funds from the Escrow, theres not much to test.

Juan-cc (approved):

salazarguille (approved):

PR #191, opened 2020-09-25
[PROTOCOL-293] TToken registry

sypeer:
It is:
- TTokenRegistry for ttokens created

It includes: 
- TTokenRegistry contract with registerTToken() and isTToken() methods 
- TTokenRegistryInterface
- Associated tests

### DIFF
--- a/contracts/base/TTokenRegistry.sol
+++ b/contracts/base/TTokenRegistry.sol
@@ -1,0 +1,94 @@
+pragma solidity 0.5.17;
+pragma experimental ABIEncoderV2;
+
+// Contracts
+import "./BaseUpgradeable.sol";
+import "./TInitializable.sol";
+
+// Interfaces
+import "../interfaces/TTokenRegistryInterface.sol";
+
+/*****************************************************************************************************/
+/**                                             WARNING                                             **/
+/**                                  THIS CONTRACT IS UPGRADEABLE!                                  **/
+/**  ---------------------------------------------------------------------------------------------  **/
+/**  Do NOT change the order of or PREPEND any storage variables to this or new versions of this    **/
+/**  contract as this will cause the the storage slots to be overwritten on the proxy contract!!    **/
+/**                                                                                                 **/
+/**  Visit https://docs.openzeppelin.com/upgrades/2.6/proxies#upgrading-via-the-proxy-pattern for   **/
+/**  more information.                                                                              **/
+/*****************************************************************************************************/
+/**
+    @notice It manages all the registred TToken contract address, mapping each one to a boolean.
+
+    @author develop@teller.finance
+ */
+contract TTokenRegistry is
+    TTokenRegistryInterface,
+    TInitializable,
+    BaseUpgradeable
+{
+
+    /* State Variables */
+
+    /**
+        @notice It represents a mapping to identify they contract address of a given TToken
+
+        i.e.:
+            address("0x479...584") => { True }.
+            address("0x132...741") => { False }.
+     */
+     mapping(address => bool) public tokenRegistry;
+
+    /**
+        @notice It registers a new TToken contract address
+        @param tToken Contract address of the TToken
+     */
+    function registerTToken(address tToken)
+        external
+        onlyPauser()
+        isInitialized()
+    {
+        _registerTTokenAddress(tToken);
+    }
+
+    /**
+        @notice Checks if a given address is a registered TToken contract
+        @param tokenAddress The token contract address to check
+        @return bool representing if the address is a registered TToken contract
+     */
+     function isTToken(address tokenAddress)
+        external
+        view
+        isInitialized()
+        returns (bool)
+    {
+        require(tokenAddress.isContract(), "INVALID_ADDRESS");
+        return(tokenRegistry[tokenAddress]);
+    }
+
+    /**
+        @notice It initializes this ttoken registry contract instance
+        @param settingsAddress The settings contract address
+     */
+     function initialize(address settingsAddress)
+        external
+        isNotInitialized() 
+    {
+        _setSettings(settingsAddress);
+    }
+
+    /** Internal functions */
+
+    /**
+        @notice It registers a new given TToken contract address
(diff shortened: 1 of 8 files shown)


contracts/base/TTokenRegistry.sol
```diff
@@ -0,0 +1,76 @@
+pragma solidity 0.5.17;
+pragma experimental ABIEncoderV2;
```
salazarguille, line 2: I think it is not required.

contracts/base/TTokenRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+        @notice It registers a new given TToken contract address
+        @param tToken Contract address of the TToken
+     */
+    function _registerTTokenAddress(address tToken) internal {
+        require(tToken.isContract(), "ADDRESS_MUST_BE_CONTRACT");
+        tokenRegistry[tToken] = true;
+
+        emit TTokenRegistered(tToken);
```
salazarguille, line 74: Please, add the msg.sender.

contracts/interfaces/TTokenRegistryInterface.sol
```diff
@@ -0,0 +1,37 @@
+pragma solidity 0.5.17;
+pragma experimental ABIEncoderV2;
```
salazarguille, line 2: I think it is not required.

salazarguille (changes requested):

contracts/base/TTokenRegistry.sol
sypeer: Removed

contracts/base/TTokenRegistry.sol
sypeer: Added

contracts/interfaces/TTokenRegistryInterface.sol
sypeer: Removed

contracts/base/TTokenRegistry.sol
```diff
@@ -0,0 +1,77 @@
...
+        _registerTTokenAddress(tToken);
+    }
+
+    /**
+        @notice Checks if a given address is a registered TToken contract
+        @param tokenAddress The token contract address to check
+        @return bool representing if the address is a registered TToken contract
+     */
+    function isTToken(address tokenAddress) external view isInitialized() returns (bool) {
+        require(tokenAddress.isContract(), "INVALID_ADDRESS");
```
salazarguille, line 51: You don't need to  add this validation.

salazarguille (changes requested):

contracts/base/TTokenRegistry.sol
sypeer: Removed

salazarguille (approved):
LGTM

passabilities (approved):

test/base/TTokenRegistryInitializeTest.js
```diff
@@ -0,0 +1,89 @@
...
+    ) {
+        it(t('user', 'initialize', 'Should (or should not) be able to initialize the TTokenRegistry contract.', mustFail), async function() {
+            // Setup
+            
+
+            const settingsAdd = contractAddress === 0 ? settingsInstance.address : NULL_ADDRESS;
+
+            try {
+                // Invocation
+                const result = await await instance.initialize(settingsAdd);;
```
Juan-cc, line 76: Do we need 2 awaits here ?

test/base/TTokenRegistryInitializeTest.js
```diff
@@ -0,0 +1,89 @@
...
+
+                // Assertions
+                assert(!mustFail, 'It should have failed because data is invalid');
+                assert(result)
```
Juan-cc, line 80: Could we check that initialization was successful? for example we could try calling a function that requires this contract to be initialized before being initialize (expecting an error), then we initialize and we successfully call the function after initialization. Or something similar that help us identify a correct state change.

test/base/TTokenRegistryInitializeTest.js
```diff
@@ -0,0 +1,89 @@
...
+
+contract('TTokenRegistryInitializeTest', function (accounts) {
+    const logicVersionsRegistryEncoder = new LogicVersionsRegistryEncoder(web3);
+    let instance;
+    let settingsInstance;
+
+    beforeEach('Setup for each test', async () => {
+        const pauser = accounts[0];
+        
+        mocks = await createMocks(Mock, 10);
```
Juan-cc, line 22: I can't find where we are using this variable?

test/base/TTokenRegistryIsTTokenTest.js
```diff
@@ -0,0 +1,94 @@
...
+        mocks = await createMocks(Mock, 10);
+        const versionsRegistryInstance = await Mock.new();
+        const constsInstance = await Mock.new();
+        await versionsRegistryInstance.givenMethodReturnAddress(
+            logicVersionsRegistryEncoder.encodeConsts(),
+            constsInstance.address
+        );
+
+        const pairAggregatorRegistryInstance = await Mock.new();
+        const settingsInstance = await createTestSettingsInstance(
```
Juan-cc, line 30: Suggestion: It would be great if we could have a function to obtain an initialized Settings instance without having to use all those params.

test/base/TTokenRegistryInitializeTest.js
sypeer: good catch, removed

test/base/TTokenRegistryInitializeTest.js
sypeer: Added an is initialized check

test/base/TTokenRegistryInitializeTest.js
sypeer: unused, removed

test/base/TTokenRegistryIsTTokenTest.js
sypeer: Agreed

PR #192, opened 2020-09-25
Fix Escrow Repay

passabilities:
It:
* Sets the allowance on the Lending Pool instead of the Loans contract
* Only takes the necessary amount of funds to repay the loan if Escrow balance does not cover the amount

### DIFF
--- a/contracts/base/Escrow.sol
+++ b/contracts/base/Escrow.sol
@@ -170,8 +170,13 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
         IERC20 token = IERC20(loans.lendingToken());
         uint256 balance = _balanceOf(loans.lendingToken());
         if (amount > balance) {
+            uint256 totalOwed = loans.getTotalOwed(loanID);
+            uint256 amountNeeded = amount > totalOwed
+                ? totalOwed.sub(balance)
+                : amount.sub(balance);
+
             require(
-                token.transferFrom(msg.sender, address(this), amount.sub(balance)),
+                token.transferFrom(msg.sender, address(this), amountNeeded),
                 "ESCROW_TRANSFER_FROM_FAILED"
             );
         }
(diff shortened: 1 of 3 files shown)


contracts/base/Escrow.sol
```diff
@@ -170,12 +170,17 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
...
         uint256 balance = _balanceOf(loans.lendingToken());
         if (amount > balance) {
+            uint256 totalOwed = loans.getTotalOwed(loanID);
+            uint256 amountNeeded = amount > totalOwed
+                ? totalOwed.sub(balance)
+                : amount.sub(balance);
+
             require(
-                token.transferFrom(msg.sender, address(this), amount.sub(balance)),
+                token.transferFrom(msg.sender, address(this), amountNeeded),
```
salazarguille, line 179: Do we need to validate the returned value?

contracts/base/Escrow.sol
```diff
@@ -170,12 +170,17 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
         IERC20 token = IERC20(loans.lendingToken());
         uint256 balance = _balanceOf(loans.lendingToken());
         if (amount > balance) {
+            uint256 totalOwed = loans.getTotalOwed(loanID);
+            uint256 amountNeeded = amount > totalOwed
```
salazarguille, line 174: I think there is a potential underflow. 
Example:
amount: 60
totalOwed: 30
balance: 40

Then the amount > balance and amount > totalOwed are true, and the substraction ```totalOwed.sub(balance)``` will fail.

salazarguille (changes requested):

salazarguille (approved):
LGTM

contracts/base/Escrow.sol
```diff
@@ -169,13 +169,18 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
     function repay(uint256 amount) external onlyOwner {
         IERC20 token = IERC20(loans.lendingToken());
```
Juan-cc, line 170: Suggestion: I would rename token to lendingToken to refer that specific token, and replace other occurrences.

contracts/base/Escrow.sol
```diff
@@ -169,13 +169,18 @@ contract Escrow is EscrowInterface, TInitializable, BaseUpgradeable, BaseEscrowD
...
         uint256 balance = _balanceOf(loans.lendingToken());
```
Juan-cc, line 171: Replace loans.lendingToken() with token (or lendingToken if previous suggestion is applied)

PR #202, opened 2020-10-12
Add: 2 ropsten signers

tpscrpt:
Adds addresses of 2 validators to ropsten config for integration testing on testnet.


salazarguille:
It was already implemented

PR #205, opened 2020-10-12
[PROTOCOL-332] Validate Signatures Using Nodes

salazarguille:
It includes:

- New SignatureValidator contract to check hashes/signatures on testnets.
- New script to request loan terms using the nodes (only testnets).


scripts/loans/createLoanWithTerms.js
```diff
@@ -57,7 +58,7 @@ module.exports = async (callback) => {
             requestNonce: nonce,
             amount: amountWithDecimals,
             duration: durationInDays * ONE_DAY,
-            requestTime: Math.round(currentTimestamp / 1000),
```
passabilities, line 60: How come this needed to change?

contracts/mock/util/SignatureValidator.sol
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity 0.5.17;
+pragma experimental ABIEncoderV2;
+
+contract SignatureValidator {
```
passabilities, line 5: We should have this contract extend the consensus contract to get all the functionality directly from them

passabilities (changes requested):

contracts/mock/util/SignatureValidator.sol
salazarguille: Fixed

scripts/loans/createLoanWithTerms.js
salazarguille: Because we changed the Timer JS class in a previous PR. Details here: https://github.com/teller-protocol/teller-protocol-v1/commit/cada826e798b43efa51d365fb2039dbafdbe595e#diff-5814c901e592fb9648cb7436c0538b7e8808f43eadd9cf2d758531dd261e5756

passabilities (approved):

PR #206, opened 2020-10-23
[PROTOCOL-344] Store cToken Info Instead of Token Info

salazarguille:
It includes:

- Modify the withdraw/deposit functions.
- Storing the cToken info in the MarketsState (if ctoken is supported)

### DIFF
--- a/contracts/base/LendingPool.sol
+++ b/contracts/base/LendingPool.sol
@@ -334,7 +334,7 @@ contract LendingPool is Base, LendingPoolInterface {
         @notice It validates whether transaction sender is the loans contract address.@
         @dev This function is overriden in some mock contracts for testing purposes.
      */
-    function _requireIsLoan() internal {
+    function _requireIsLoan() internal view {
         loans.requireEqualTo(msg.sender, "ADDRESS_ISNT_LOANS_CONTRACT");
     }
 
--- a/contracts/base/LendingPool.sol
+++ b/contracts/base/LendingPool.sol
@@ -228,7 +228,6 @@ contract LendingPool is Base, LendingPoolInterface {
 
         // Withdraw the tokens from compound
         _withdrawFromCompoundIfSupported(amountToWithdraw);
-        // TODO Do we need to decrease any amount?
 
         // Transfer tokens to the lender.
         tokenTransfer(lender, amountToWithdraw);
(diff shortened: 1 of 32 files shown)


contracts/base/MarketsState.sol
```diff
@@ -61,7 +70,12 @@ contract MarketsState is
         address collateralAsset,
         uint256 amount
     ) external onlyWhitelisted() isInitialized() {
-        markets[borrowedAsset][collateralAsset].increaseRepayment(amount);
+        address cTokenAddress = _getCTokenAddress(borrowedAsset);
+        if (cTokenAddress.isEmpty()) {
+            markets[borrowedAsset][collateralAsset].increaseRepayment(amount);
+        } else {
+            markets[cTokenAddress][collateralAsset].increaseRepayment(amount);
+        }
```
passabilities, line 78: Because the same logic is duplicated several times in this file, we should instead change the `_getCTokeAddress` function to `_getMarket` and just use that to perform the action needed in each function.

passabilities (changes requested):

contracts/base/LendingPool.sol
```diff
@@ -105,15 +106,15 @@ contract LendingPool is Base, LendingPoolInterface {
...
-        _withdrawFromCompoundIfSupported(amount);
+        uint256 withdrawnAmount = _withdrawFromCompoundIfSupported(amount);
 
         // Transfers tokens
         tokenTransfer(msg.sender, amount);
 
         _markets().decreaseSupply(
             address(lendingToken),
             LoansInterface(loans).collateralToken(),
-            amount
```
Juan-cc, line 116: Should we include amount value in the event also? (for tracking purposes) So we know what amount was requested, and what actually was withdrawn.

contracts/base/LendingPool.sol
```diff
@@ -215,13 +228,22 @@ contract LendingPool is Base, LendingPoolInterface {
 
         // Withdraw the tokens from compound
         _withdrawFromCompoundIfSupported(amountToWithdraw);
+        // TODO Do we need to decrease any amount?
```
Juan-cc, line 231: Please implement this TODO before merging.

Juan-cc (changes requested):

contracts/base/LendingPool.sol
salazarguille: It makes sense to me. Just to keep the scope of this task, I created a new task to implement your change.

contracts/base/LendingPool.sol
salazarguille: I created a task to analyze/implement. Now we don't need to get/calculate the total interest withdrawn.

passabilities (approved):

Juan-cc (approved):

sypeer (approved):

PR #208, opened 2020-10-30
[PROTOCOL-296] Enable Over Collateralized Loans without Escrow

passabilities:
It:
* Calculates the minimum amount of collateral ratio required for a loan to go to user's wallet

Creating an integration test for this and will push in another commit.


contracts/util/SettingsConsts.sol
```diff
@@ -12,6 +12,9 @@ pragma solidity 0.5.17;
 contract SettingsConsts {
     /** Constants */
 
+    // It represents 100% with 2 decimal places.
+    uint256 public constant TEN_THOUSAND = 10000;
```
Juan-cc, line 16: I think it would help to understand the constant if we change the constant name with the corresponding meaning instead of value. Something like HUNDRED_PERCENT or similar

Juan-cc (approved):

salazarguille (approved):

sypeer (approved):

PR #210, opened 2020-11-04
[PROTOCOL-336] Update getTotalOwed function

sypeer:
It is:
- Updated getTotalOwed function
- Updated _getCollateralNeededInTokens
 
Screen Shot 2020-11-04 at 5 35 07 PM


contracts/base/LoansBase.sol
```diff
@@ -568,18 +568,31 @@ contract LoansBase is LoansInterface, Base {
...
+     function _getTotalOwed(uint256 loanID)
+        internal
+        view
+        returns (uint256)
     {
         TellerCommon.LoanStatus currentStatus = loans[loanID].status;
         uint256 loanAmount;
         if (currentStatus == TellerCommon.LoanStatus.TermsSet) {
-            loanAmount = loans[loanID].loanTerms.maxLoanAmount;
+            loanAmount = loans[loanID].loanTerms.maxLoanAmount.add(loans[loanID].interestOwed);
```
passabilities, line 589: Interest owed won't be set yet if the loan is not yet active. So we'll have to calculate the interest based on the max loan amount. 

Take the calculation code for the interest that is in `takeOutLoan` here:
https://github.com/teller-protocol/teller-protocol-v1/blob/72a6d1d17f77a1226a5f3a1855ae1fec2e02da04/contracts/base/LoansBase.sol#L254-L258
and maybe put it in its own internal function to reuse.

passabilities (changes requested):
Also write a simple unit test/integration test for this to check the values being calculated.

contracts/base/LoansBase.sol
sypeer: Changed accordingly

contracts/base/LoansBase.sol
```diff
@@ -586,7 +582,8 @@ contract LoansBase is LoansInterface, Base {
         TellerCommon.LoanStatus currentStatus = loans[loanID].status;
         uint256 loanAmount;
         if (currentStatus == TellerCommon.LoanStatus.TermsSet) {
-            loanAmount = loans[loanID].loanTerms.maxLoanAmount.add(loans[loanID].interestOwed);
+            uint256 interestOwed = _getInterestOwed(loanID, loans[loanID].loanTerms.maxLoanAmount);
+            loanAmount = loans[loanID].loanTerms.maxLoanAmount.add(interestOwed);
         } else if (currentStatus == TellerCommon.LoanStatus.Active) {
             loanAmount = loans[loanID].interestOwed.add(loans[loanID].principalOwed);
```
passabilities, line 588: Just a preference to have the principalOwed before the interest 😅

contracts/base/LoansBase.sol
```diff
@@ -595,6 +592,24 @@ contract LoansBase is LoansInterface, Base {
...
+        internal
+        view
+        returns (uint256)
+    {
+        uint256 interestOwed = amountBorrow
+            .mul(loans[loanID].loanTerms.interestRate)
+            .mul(loans[loanID].loanTerms.duration)
+            .div(TEN_THOUSAND)
+            .div(SECONDS_PER_YEAR_4DP);
+        return interestOwed;
```
passabilities, line 610: lets just return the value instead of creating a variable

passabilities (changes requested):

contracts/base/LoansBase.sol
```diff
@@ -568,18 +564,50 @@ contract LoansBase is LoansInterface, Base {
...
         if (currentStatus == TellerCommon.LoanStatus.TermsSet) {
-            loanAmount = loans[loanID].loanTerms.maxLoanAmount;
+            uint256 interestOwed = _getInterestOwed(loanID, loans[loanID].loanTerms.maxLoanAmount);
+            loanAmount = loans[loanID].loanTerms.maxLoanAmount.add(interestOwed);
         } else if (currentStatus == TellerCommon.LoanStatus.Active) {
-            loanAmount = getTotalOwed(loanID);
+            loanAmount = loans[loanID].interestOwed.add(loans[loanID].principalOwed);
         } else {
-            return 0;
+            loanAmount = 0;
```
passabilities, line 590: also here you can just return 0. no need to set the variable

contracts/base/LoansBase.sol
sypeer: Got it, changed

contracts/base/LoansBase.sol
sypeer: Done

contracts/base/LoansBase.sol
sypeer: Done

passabilities (approved):

salazarguille (approved):

PR #211, opened 2020-11-05
[PROTOCOL-379] Escrow: Removing Library & Fixing StD Ratio

salazarguille:
It includes:

- Remove library SignedSafeMath from Escrow.
- New signers.
- New value for SafetyInterval in Ropsten.
- Fix to StD ratio.


contracts/providers/compound/CErc20Interface.sol
```diff
@@ -72,6 +72,8 @@ interface CErc20Interface {
      */
     function exchangeRateCurrent() external view returns (uint256);
```
passabilities, line 73: we should remove the `view` modifier since it is actually not a view funciton.

passabilities (changes requested):

test-integration/scenarios/loans/1_eth_takeout_loan_and_2_repays_full_successfully.js
```diff
@@ -1,153 +0,0 @@
-// Util classes
```
passabilities, line 1: why are we deleting this?

salazarguille: We already have this scenario in file ```test-integration/scenarios/loans/1_takeout_loan_and_2_repays_full_successfully.js``` (and it is supporting the multi execution per market)

sypeer (approved):

passabilities (approved):

PR #218, opened 2020-11-13
[PROTOCOL-381] Use SafeERC20 for handling tokens

sypeer:
It is:
- Implementing use of SafeERC20 library for external tokens used in the protocol to correctly handle all possible incorrect implementations of ERC20

Screen Shot 2020-11-12 at 7 34 15 PM


contracts/base/BaseEscrowDapp.sol
```diff
@@ -2,16 +2,15 @@ pragma solidity 0.5.17;
...
 
-// Interfaces
-import "@openzeppelin/contracts-ethereum-package/contracts/token/ERC20/IERC20.sol";
-
 // Libraries
 import "../util/AddressArrayLib.sol";
 
 contract BaseEscrowDapp is Ownable, BaseUpgradeable {
     using AddressArrayLib for address[];
+    using SafeERC20 for IERC20;
```
hensha256, line 13: You never use any of the SafeERC20 functions in this contract. I dont think its needed :)

sypeer: Ah yes, the Uniswap and Compound apps inherit from this base escrow, which is why this was here. I've imported SafeERC20 into each of them and removed this.

passabilities (approved):

PR #220, opened 2020-11-14
A few tiny changes

hensha256:


sypeer (approved):

contracts/util/AddressArrayLib.sol
```diff
@@ -61,7 +56,7 @@ library AddressArrayLib {
         view
         returns (bool found, uint256 indexAt)
     {
-        found = false;
+        found;
```
passabilities, line 59: can probably just delete this line

passabilities (changes requested):

contracts/util/AddressArrayLib.sol
hensha256: oh yeah good point 👍

passabilities (approved):

PR #224, opened 2020-11-21
Extract loan functions to LoanLib library.

passabilities:
Screen Shot 2020-12-06 at 10 39 48 PM


sypeer (approved):

contracts/base/LoansBase.sol
```diff
@@ -109,7 +103,7 @@ contract LoansBase is LoansInterface, Base {
         @param loanID number of loan to check
      */
     modifier loanActiveOrSet(uint256 loanID) {
-        require(_isLoanActiveOrSet(loanID), "LOAN_NOT_ACTIVE_OR_SET");
+        require(loans[loanID].isActiveOrSet(), "LOAN_NOT_ACTIVE_OR_SET");
```
salazarguille, line 106: We also might to add a requireIsActiveOrSet() function in the library.

contracts/base/LoansBase.sol
```diff
@@ -170,22 +164,93 @@ contract LoansBase is LoansInterface, Base {
...
-        return _isLoanSecured(loanID);
+        return loans[loanID].isSecured(settings());
     }
 
-    /**
-        @notice Checks whether a loan is allowed to be deposited to an Externally Owned Account.
-        @param loanID Id of the loan to check.
-        @return bool indicating whether the loan with specified parameters can be deposited to an EOA.
-     */
+    // See LoanLib.canGoToEOA
```
salazarguille, line 172: I would add @dev See more details in LoanLib.canGoToEOA(...) function.
Same in other functions.

contracts/util/LoanLib.sol
```diff
@@ -0,0 +1,339 @@
...
+import "../interfaces/SettingsInterface.sol";
+import "../interfaces/EscrowInterface.sol";
+import "../interfaces/LoansInterface.sol";
+
+// Contracts
+
+library LoanLib {
+    using SafeMath for uint256;
+    using NumbersLib for uint256;
+    using NumbersLib for int256;
```
salazarguille, line 19: I think NumbersLib is defined for uint256 (not int256).

contracts/util/LoanLib.sol
```diff
@@ -0,0 +1,339 @@
...
+                collateral: loan.collateral,
+                valueInLendingTokens: getCollateralInLendingTokens(loan, loansContract),
+                escrowLoanValue: escrowLoanValue,
+                neededInLendingTokens: neededInLending,
+                neededInCollateralTokens: neededInCollateral,
+                moreCollateralRequired: neededInCollateral > int256(loan.collateral)
+            });
+    }
+
+    function getCollateralInLendingTokens(
```
salazarguille, line 177: Please, add docs.

contracts/util/LoanLib.sol
```diff
@@ -0,0 +1,339 @@
...
+        returns (bool)
+    {
+        return
+            loan.loanTerms.collateralRatio >=
+            settings.getPlatformSettingValue(
+                settings.consts().COLLATERAL_BUFFER_SETTING()
+            );
+    }
+
+    function isActiveOrSet(TellerCommon.Loan storage loan) public view returns (bool) {
```
salazarguille, line 113: Please, add docs.

contracts/util/LoanLib.sol
```diff
@@ -0,0 +1,339 @@
...
+     */
+    function getLiquidationInfo(
+        TellerCommon.Loan storage loan,
+        LoansInterface loansContract
+    ) public view returns (TellerCommon.LoanLiquidationInfo memory liquidationInfo) {
+        liquidationInfo.collateralInfo = getCollateralInfo(loan, loansContract);
+        liquidationInfo.amountToLiquidate = getTotalOwed(loan);
+
+        // Maximum reward is the calculated value of required collateral minus the principal owed (see LoanLib.getCollateralNeededInTokens).
+        int256 maxReward = liquidationInfo.collateralInfo.neededInLendingTokens -
```
salazarguille, line 300: why we are using int256 (instead of uint256)? Please, use `value.sub(anotherValue)` instead of `-`

contracts/util/LoanLib.sol
```diff
@@ -0,0 +1,339 @@
...
+            toPay = toPay.sub(loan.interestOwed);
+            loan.interestOwed = 0;
+            if (toPay == 0) {
+                return;
+            }
+        } else {
+            loan.interestOwed = loan.interestOwed.sub(toPay);
+            return;
+        }
+        loan.principalOwed = loan.principalOwed.sub(toPay);
```
salazarguille, line 337: Since the variables used in line 337 are not used in the "if/else" statements, it might be moved as the first line in the function.
So then, we might remove the `return;` in the else (and the "toPay == 0" condition too).

```
        loan.principalOwed = loan.principalOwed.sub(toPay);
        if (toPay >= loan.interestOwed) {
            toPay = toPay.sub(loan.interestOwed);
            loan.interestOwed = 0;
        } else {
            loan.interestOwed = loan.interestOwed.sub(toPay);
        }
```

What would happen if "toPay > principalOwed". Is it possible?

migrations/3_deploy.js
```diff
@@ -100,6 +102,10 @@ module.exports = async function(deployer, network, accounts) {
       { Contract: TTokenRegistry, name : logicNames.TTokenRegistry },
     ];
 
+    const loanLib = await LoanLib.new();
```
salazarguille, line 105: Why don't we use the deployerApp.deploy(...) for libraries?

Also, there is a `deployerApp.links(contract, libraries)` function. So, the deployer app stores the addresses (including the library)

test/base/EscrowClaimTokensTest.js
```diff
@@ -14,29 +14,32 @@ const DAI = artifacts.require("./mock/token/DAIMock.sol");
...
-    _4_loan_liquidated_recipient_not_borrower_caller_loans: [ loanStatus.Closed, accounts[1], accounts[2], true, true, 2, 1000, false, null ],
-    _5_loan_liquidated_recipient_is_borrower_caller_loans: [ loanStatus.Closed, accounts[1], accounts[1], true, true, 2, 1000, false, null ],
-    _6_loan_liquidated_recipient_not_borrower_caller_not_loans: [ loanStatus.Closed, accounts[1], accounts[2], false, true, 2, 1000, true, "CALLER_MUST_BE_LOANS" ],
+    _1_loan_active: [ loanStatus.Active, accounts[1], accounts[1], false, 0, 0, true, "LOAN_NOT_CLOSED" ],
+    _2_loan_not_liquidated_recipient_not_borrower: [ loanStatus.Closed, accounts[1], accounts[2], false, 0, 0, true, "LOAN_NOT_LIQUIDATED" ],
+    _3_loan_not_liquidated_recipient_is_borrower: [ loanStatus.Closed, accounts[1], accounts[1], false, 2, 1000, true, "LOAN_NOT_LIQUIDATED" ],
+    _4_loan_liquidated_recipient_is_borrower: [ loanStatus.Closed, accounts[1], accounts[1], true, 2, 1000, false, null ],
+    _5_loan_liquidated_recipient_not_borrower: [ loanStatus.Closed, accounts[1], accounts[2], true, 2, 1000, true, "CALLER_MUST_BE_LOANS" ],
```
salazarguille, line 38: Why we removed a scenario?

test/base/EscrowFactoryCreateEscrowTest.js
```diff
@@ -26,6 +29,8 @@ contract("EscrowFactoryCreateEscrowTest", function(accounts) {
 
   beforeEach(async () => {
     const CETH = await Mock.new();
+    const loanLib = await LoanLib.new();
```
salazarguille, line 32: This is painful.

test/base/LoansBaseIsLoanSecuredTest.js
```diff
@@ -12,6 +12,9 @@ const Mock = artifacts.require("./mock/util/Mock.sol");
 const Loans = artifacts.require("./mock/base/LoansBaseMock.sol");
 const Settings = artifacts.require("./base/Settings.sol");
```
salazarguille, line 13: Why the file test/base/LoansBaseGetCollateralInfoTest.js is empty?

salazarguille (changes requested):

salazarguille:
Great job Noah!
Please, add a description in the PR. Are the integration tests passing? Can you add a screenshot?

contracts/base/LoanTermsConsensus.sol
```diff
@@ -96,6 +96,7 @@ contract LoanTermsConsensus is LoanTermsConsensusInterface, Consensus {
         );
         borrowerToLastLoanTermRequest[request.borrower] = now;
 
+        // TODO: Remove redundant event
```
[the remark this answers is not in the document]
passabilities: because the only time these events are emitted are in the same transaction:
https://github.com/teller-protocol/teller-protocol-v1/blob/dd0704e9736a940154c45113b202d6842ad83310/contracts/base/LoansBase.sol#L244-L253

contracts/util/LoanLib.sol
passabilities: Some functions can be used for both

contracts/util/LoanLib.sol
passabilities: We spend a good minute (couple hours) going over this. Basically the reward and the required collateral is adjusted based on the value of the escrow. 

The equation for being able to liquidate is:
`canLiquidate(collateralAmount) = collateralValue < owed + liquidation reward - escrow value`

The max liquidation reward is 5% of what the liquidator pays on whats owed for the loan. Now, if the value of the collateral and escrow drop below the 5% value, we cannot pay the liquidator the 5%. 
AND 
If the value of the collateral and escrow continue to drop, there could potentially be a negative reward amount. A.K.A. the liquidator pays the loan and receives less value than he paid. Unlikely to happen, but possible.

Same thing with the `collateralInfo.neededInLendingTokens`. If the collateral and escrow value rise in value, you are not required to have as much locked in collateral. If the collateral and escrow grow past a certain point, the `neededInLendingTokens` can become negative. That basically means that the borrower can withdraw the full collateral value but nothing else.

contracts/util/LoanLib.sol
passabilities: Moving the principal to the bottom allows us to pay back the interest first. This is how traditional loans work, but also the required collateral amount is calculated by subtracting values from the collateral ratio. The collateral ratio that is defined on the loan struct is the required amount to take out the loan but once the loan is taken out we can subtract percentage values to know that the min collateral amount is to keep the loan in good standing.

We can calculate the min collateral required by only using the `principalOwed` value because:
1. We pay off the interest first
1. We subtract the interest rate from the collateral ratio

https://github.com/teller-protocol/teller-protocol-v1/blob/dd0704e9736a940154c45113b202d6842ad83310/contracts/util/LoanLib.sol#L261
https://github.com/teller-protocol/teller-protocol-v1/blob/dd0704e9736a940154c45113b202d6842ad83310/contracts/util/LoanLib.sol#L272-L275

test/base/EscrowFactoryCreateEscrowTest.js
passabilities: Yes, but only solid way to reduce the size of the LoansBase contract, unfortunately.

test/base/EscrowClaimTokensTest.js
passabilities: This function was split into 2 on the Escrow. 1 for paying out the liquidator and the other for the borrower to claim the remaining tokens.

test/base/LoansBaseIsLoanSecuredTest.js
passabilities: Should not be any more. We (@sypeer) were fixing the test

salazarguille (changes requested):

contracts/base/LoansBase.sol
sypeer: Added

contracts/util/LoanLib.sol
sypeer: Added

contracts/util/LoanLib.sol
sypeer: Added

PR #232, opened 2020-12-07
[PROTOCOL-407] Track loans with userId

tpscrpt:
userId is a field provided in the loanRequest and verified through the standard loanTermsConsensus hashing/signature verification mechanism. It is generated in the centralized data-provider based on the user's identifiable details.

We ensure that a provided userId is not used to take out more than 1 unsecured loan (< 30% collateral).

We do this by mapping userIds to a boolean (could be changed to a limit on number of loans or total lent in the future) and requiring that this value be false for a given loanRequest in withValidLoanRequest.

The value is set to `true` for a given userId at the end of createLoanWithTerms and back to `false` when repay leaves a loan with no outstanding balance.

EDIT: if a user doesn't provide bank info on the validator side, the userId is set to bytes32(0).


contracts/base/Escrow.sol
```diff
@@ -153,7 +153,10 @@ contract Escrow is EscrowInterface, TInitializable, BaseEscrowDapp {
         @dev The recipient must be the loan borrower AND the loan must be already liquidated.
     */
     function claimTokens() external onlyOwner() {
-        require(getLoan().status == TellerCommon.LoanStatus.Closed, "LOAN_NOT_CLOSED");
+        require(
+            getLoan().status == TellerCommon.LoanStatus.Liquidated,
```
passabilities, line 157: Because this function can only be called by the owner, we should not be checking that the loan is `Liquidated` but rather just `Closed`. If a user pays back their loan, we still want them to be able to claim their tokens that are still in the Escrow.

test/base/EscrowClaimTokensTest.js
```diff
@@ -31,24 +31,23 @@ contract("EscrowClaimTokensTest", function(accounts) {
...
-    _2_loan_not_liquidated_recipient_not_borrower: [ loanStatus.Closed, accounts[1], accounts[2], false, 0, 0, true, "LOAN_NOT_LIQUIDATED" ],
-    _3_loan_not_liquidated_recipient_is_borrower: [ loanStatus.Closed, accounts[1], accounts[1], false, 2, 1000, true, "LOAN_NOT_LIQUIDATED" ],
-    _4_loan_liquidated_recipient_is_borrower: [ loanStatus.Closed, accounts[1], accounts[1], true, 2, 1000, false, null ],
-    _5_loan_liquidated_recipient_not_borrower: [ loanStatus.Closed, accounts[1], accounts[2], true, 2, 1000, true, "CALLER_MUST_BE_LOANS" ],
+    _1_loan_active: [ loanStatus.Active, accounts[1], accounts[1], 0, 0, true, "LOAN_NOT_LIQUIDATED" ],
+    _2_loan_not_liquidated_recipient_not_borrower: [ loanStatus.Closed, accounts[1], accounts[2], 0, 0, true, "LOAN_NOT_LIQUIDATED" ],
+    _3_loan_not_liquidated_recipient_is_borrower: [ loanStatus.Closed, accounts[1], accounts[1], 2, 1000, true, "LOAN_NOT_LIQUIDATED" ],
+    _4_loan_liquidated_recipient_is_borrower: [ loanStatus.Liquidated, accounts[1], accounts[1], 2, 1000, false, null ],
+    _5_loan_liquidated_recipient_not_borrower: [ loanStatus.Liquidated, accounts[1], accounts[2], 2, 1000, true, "CALLER_MUST_BE_LOANS" ],
```
passabilities, line 38: No need to be checking that the loan is `Liquidated` 

1. Should throw when loan is active
1. Should throw when caller not the owner
1. Should only pass when loan is `Closed` and caller is owner

test/base/EstimateGasEscrowClaimTokensTest.js
```diff
@@ -17,8 +17,8 @@ const Escrow = artifacts.require("./mock/base/EscrowMock.sol");
 contract("EstimateGasEscrowClaimTokensTest", function(accounts) {
   const loansEncoder = new LoansBaseInterfaceEncoder(web3);
   
-  const baseGasCost = 452022; // Gas cost with 1 token in wallet
-  const expectedGasCost = (tokens) => baseGasCost + ((tokens -  1) * 37000); // Gas cost > 1 token in wallet
+  const baseGasCost = 453453; // Gas cost with 1 token in wallet
+  const expectedGasCost = (tokens) => baseGasCost + ((tokens -  1) * 167151); // Gas cost > 1 token in wallet
```
passabilities, line 21: this is quite a hike in expected gas cost @tpscrpt @sypeer

test/base/EstimateGasEscrowClaimTokensTest.js
```diff
@@ -48,7 +48,7 @@ contract("EstimateGasEscrowClaimTokensTest", function(accounts) {
     it(t("escrow", "claimTokens", "Should be able to claim all tokens in the escrow", false), async function() {
       await loans.givenMethodReturn(
         loansEncoder.encodeLoans(),
-        encodeLoanParameter(web3, { status: Closed, liquidated: true, loanTerms: { borrower } })
+        encodeLoanParameter(web3, { status: Liquidated, loanTerms: { borrower } })
```
passabilities, line 51: Should just be `Closed`

test/base/EtherCollateralLoansGetCollateralInfoTest.js
```diff
@@ -120,6 +121,8 @@ contract("EtherCollateralLoansGetCollateralInfoTest", function(accounts) {
       let collateralNeededInTokens;
       switch (status) {
         case NON_EXISTENT:
+        case LIQUIDATED:
+          collateralNeededInTokens = new BigNumber(0);
         case CLOSED:
           collateralNeededInTokens = new BigNumber(0);
```
passabilities, line 127: Can remove line 125 and just let the switch statement fall through

scripts/utils/forkRopsten.js
```diff
@@ -0,0 +1,46 @@
+const Ganache = require("ganache-core");
```
passabilities, line 1: Lets add a CLI param to pass in the network flag for the existing `forkMainnet.js`

contracts/base/LoansBase.sol
```diff
@@ -241,6 +247,10 @@ contract LoansBase is LoansInterface, Base {
 
         borrowerLoans[request.borrower].push(loanID);
 
+        if (request.userId != bytes32(0)) {
+            userIdHasLoan[request.userId] = true;
+        }
```
passabilities, line 252: We'll need to validate that this flag is not already set OR if it is check their previous unsecured loan status. If its `TermsSet`, meaning it was never taken out, validate that the terms have expired.

Also I do not see a check on the `collateralRatio` value @tpscrpt this should only happen when the collateral ratio is 0, right?

contracts/base/LoansBase.sol
```diff
@@ -420,6 +430,11 @@ contract LoansBase is LoansInterface, Base {
         if (totalOwed == 0) {
             loans[loanID].status = TellerCommon.LoanStatus.Closed;
 
+            bytes32 userId = loans[loanID].userId;
+            if (userId != bytes32(0)) {
+                userIdHasLoan[userId] = false;
+            }
```
passabilities, line 436: We'll need to put this when a loan is liquidated as well since the loan is also repaid their.

passabilities (changes requested):

test/base/EstimateGasEscrowClaimTokensTest.js
tpscrpt: Yeah I'll review this again and try to find why, currently my best guess is just the added bytes(32) read from storage on every getLoan() call (since it's kept in another contract and can't just extract 1 value). Might have to refactor so this loan gets passed around instead of read multiple times from storage.

tpscrpt:
@passabilities https://github.com/teller-protocol/teller-protocol-v1/pull/232/commits/f7333988263dee5332be687435d417060ad70838

Addresses some concerns with TermsSet loans, read the commit message and code diff for context

contracts/base/LoansBase.sol
```diff
@@ -383,6 +396,10 @@ contract LoansBase is LoansInterface, Base {
             EscrowInterface(loans[loanID].escrow).initialize(address(this), loanID);
         }
 
+        if (loans[loanID].userId != bytes32(0)) {
+            userIdHasLoan[loans[loanID].userId] = true;
+        }
```
passabilities, line 401: I assume that because the `loans[loanID].userId` value is set it meas that the loan is unsecured? Lets also add the check `!loans[loanID].isLoanSecured()` which is the solidity implementation.

contracts/base/LoansBase.sol
```diff
@@ -626,4 +646,15 @@ contract LoansBase is LoansInterface, Base {
     function _createEscrow(uint256 loanID) internal returns (address) {
         return settings().escrowFactory().createEscrow(address(this), loanID);
     }
+
+    /**
+        @notice If the loan was tied to bank info (userId provided), reset the userId.
+        @param loanID loan id associated to the Escrow contract.
+     */
+    function _clearUserIdLoan(uint256 loanID) internal {
+        bytes32 userId = loans[loanID].userId;
```
passabilities, line 655: No need for local variable here; just adds more gas cost

passabilities (changes requested):

contracts/base/LoansBase.sol
```diff
@@ -626,4 +646,15 @@ contract LoansBase is LoansInterface, Base {
...
+        if (userId != bytes32(0)) {
```
passabilities, line 656: Also we need to be checking that the `loanID` here is unsecured (`!loans[loanID].isSecured()`) before we clear the boolean

passabilities (changes requested):

PR #235, opened 2020-12-13
[PROTOCOL-402] Create Settings for Max Total Value Locked

sypeer:
It is:
- Created settings for the max locked value allowed.
- Add require statement in LendingPool.deposit

Screen Shot 2020-12-13 at 6 38 23 PM

### DIFF
--- a/config/networks/test/platformSettings.json
+++ b/config/networks/test/platformSettings.json
@@ -53,6 +53,12 @@
     "min": 11000,
     "max": 50000
   },
+  "MaximumTotalValueLocked": {
+    "processOnDeployment": true,
+    "value": 2000,
+    "min": 0,
+    "max": 2000
+  },
   "StartingBlockOffsetNumber": {
     "processOnDeployment": true,
     "value": 40,
--- a/config/networks/test/platformSettings.json
+++ b/config/networks/test/platformSettings.json
@@ -55,9 +55,9 @@
   },
   "MaximumTotalValueLocked": {
     "processOnDeployment": true,
-    "value": 2000,
+    "value": 100000,
     "min": 0,
-    "max": 2000
+    "max": 100000
   },
   "StartingBlockOffsetNumber": {
     "processOnDeployment": true,
--- a/config/networks/test/platformSettings.json
+++ b/config/networks/test/platformSettings.json
@@ -53,11 +53,11 @@
     "min": 11000,
     "max": 50000
   },
-  "MaximumTotalValueLocked": {
+  "PoolMaximumTotalValueLocked": {
     "processOnDeployment": true,
     "value": 100000,
     "min": 0,
-    "max": 100000
+    "max": "115792089237316195423570985008687907853269984665640564039457584007913129639935"
   },
   "StartingBlockOffsetNumber": {
     "processOnDeployment": true,
--- a/contracts/base/LendingPool.sol
+++ b/contracts/base/LendingPool.sol
@@ -76,7 +76,7 @@ contract LendingPool is Base, LendingPoolInterface {
             _getSettings().consts().MAXIMUM_TOTAL_VALUE_LOCKED()
         );
         require(
-            lendingToken.balanceOf(address(this)) <= maxTotalValueLocked,
+            lendingToken.balanceOf(address(this)).add(amount) <= maxTotalValueLocked,
             "TVL_MAXED"
         );
         // Transfering tokens to the LendingPool
--- a/contracts/base/LendingPool.sol
+++ b/contracts/base/LendingPool.sol
@@ -73,7 +73,7 @@ contract LendingPool is Base, LendingPoolInterface {
         whenLendingPoolNotPaused(address(this))
     {
         uint256 maxTotalValueLocked = _getSettings().getPlatformSettingValue(
-            _getSettings().consts().MAXIMUM_TOTAL_VALUE_LOCKED()
+            _getSettings().consts().POOL_MAXIMUM_TOTAL_VALUE_LOCKED()
         );
         require(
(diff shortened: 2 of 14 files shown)


contracts/base/LendingPool.sol
```diff
@@ -72,6 +72,13 @@ contract LendingPool is Base, LendingPoolInterface {
         whenNotPaused()
         whenLendingPoolNotPaused(address(this))
     {
+        uint256 maxTotalValueLocked = _getSettings().getPlatformSettingValue(
+            _getSettings().consts().MAXIMUM_TOTAL_VALUE_LOCKED()
+        );
+        require(
+            lendingToken.balanceOf(address(this)).add(amount) <= maxTotalValueLocked,
+            "TVL_MAXED"
+        );
```
passabilities, line 81: this only works when the lending token is not deposited to Compound. Instead, for now, we should be using the MarketState value. 

Also, because we are just launching with one market, this is fine, but if we were to have more markets, we would want to convert the value of the asset into 1 of the "USD value" tokens. I.e. convert DAI into USDC and use USDC as the normalized value for the max cap.

passabilities (changes requested):

contracts/util/SettingsConsts.sol
```diff
@@ -78,4 +78,9 @@ contract SettingsConsts {
      */
     bytes32
         public constant OVER_COLLATERALIZED_BUFFER_SETTING = "OverCollateralizedBuffer";
+    /**
+        @notice The setting name for the maximum total value locked.
+        @notice The maximum total value locked is the supply cap for protocol, above which no additional deposits will be accepted.
```
Juan-cc, line 83: Suggestion:        @notice The maximum total value locked is the supply cap for a pool protocol, above which no additional deposits will be accepted on this pool.

contracts/util/SettingsConsts.sol
```diff
@@ -78,4 +78,9 @@ contract SettingsConsts {
...
+     */
+    bytes32 public constant MAXIMUM_TOTAL_VALUE_LOCKED = "MaximumTotalValueLocked";
```
Juan-cc, line 85: probably if we make it explicitly related to the pool would be easier to understand, something like: -POOL_MAXIMUM_TOTAL_VALUE_LOCKED instead of MAXIMUM_TOTAL_VALUE_LOCKED

Juan-cc (approved):

contracts/base/LendingPool.sol
```diff
@@ -72,6 +72,13 @@ contract LendingPool is Base, LendingPoolInterface {
         whenNotPaused()
         whenLendingPoolNotPaused(address(this))
     {
+        uint256 maxTotalValueLocked = _getSettings().getPlatformSettingValue(
+            _getSettings().consts().MAXIMUM_TOTAL_VALUE_LOCKED()
```
salazarguille, line 76: As Juan mentioned in the other contract. I would change the constant name to refer to the max TLV for the pool (it looks like it refers to the TVL for the platform).

test/base/LendingPoolDepositTest.js
```diff
@@ -36,25 +37,42 @@ contract('LendingPoolDepositTest', function (accounts) {
...
+                initialize: true,
+                onInitialize: async(instance, { marketsState }) => {
+                    marketsInstance = marketsState;
+                }});
+          
+        await settingsInstance.createAssetSettings(daiInstance.address, cTokenInstance.address, 1000, { from: accounts[0] });
+
+        await settingsInstance.createPlatformSetting(
+            toBytes32(web3, settingsNames.MaximumTotalValueLocked),
+            2000,
```
salazarguille, line 70: Should we consider the decimals in the calculation to get the max TVL?. There are no issues when there is only one market, but what will happen with two markets. i.e., DAI/ETH and USDC/ETH.

I.e., USDC has 6 decimals, and DAI has 18 decimals.

P.S.: There is an AssetSettings

test/base/LendingPoolWithdrawTest.js
```diff
@@ -118,6 +121,13 @@ contract('LendingPoolWithdrawTest', function (accounts) {
     }, function(depositSender, depositAmount, recipient, amountToWithdraw, expectedErrorMessage, mustFail) {
         it(t('user', 'withdraw', 'Should able (or not) to withdraw DAIs.', mustFail), async function() {
             // Setup
+            await settingsInstance.createPlatformSetting(
+                toBytes32(web3, settingsNames.MaximumTotalValueLocked),
+                2000,
```
salazarguille, line 126: The same here

salazarguille (changes requested):

config/networks/test/platformSettings.json
```diff
@@ -53,6 +53,12 @@
     "min": 11000,
     "max": 50000
   },
+  "MaximumTotalValueLocked": {
+    "processOnDeployment": true,
+    "value": 100000,
+    "min": 0,
+    "max": 100000
+  },
```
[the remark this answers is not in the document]
sypeer: Done

contracts/util/SettingsConsts.sol
sypeer: Added in

contracts/util/SettingsConsts.sol
sypeer: Changed

contracts/base/LendingPool.sol
sypeer: Changed

contracts/base/LendingPool.sol
salazarguille: Also, I think the setting is not considering the decimals.

salazarguille (changes requested):

PR #247, opened 2020-12-30
[PROTOCOL-414] [L03] Support Multiple Versions for a Logic Address

salazarguille:
It includes:

- Changes to support multiple versions for a logic version.
- Delete the remove function (comment in PR).

### DIFF
--- a/contracts/util/LogicVersionLib.sol
+++ b/contracts/util/LogicVersionLib.sol
@@ -60,7 +60,7 @@ library LogicVersionLib {
         currentVersion = self.currentVersion;
         previousLogic = self.logicVersions[self.currentVersion];
         newLogic = self.logicVersions[previousVersion];
-        
+
         self.currentVersion = previousVersion;
     }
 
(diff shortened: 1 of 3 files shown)


contracts/util/LogicVersionLib.sol
```diff
@@ -93,10 +124,13 @@ library LogicVersionLib {
...
-        self.version = 0;
+        lastLogic = self.logicVersions[self.latestVersion];
+        lastVersion = self.latestVersion;
+
+        self.logicVersions[self.currentVersion] = address(0x0);
+        self.logicVersions[self.latestVersion] = address(0x0);
+        self.currentVersion = 0;
+        self.latestVersion = 0;
         self.exists = false;
     }
```
passabilities, line 135: I want to say that we should just remove the functionality to remove a logic version. I see no reason to have this functionality

passabilities (changes requested):
Looks great! Just one suggestion below

passabilities (approved):

PR #253, opened 2021-01-04
[PROTOCOL-434] Separate Interest Payments

hensha256:


contracts/util/LoanLib.sol
```diff
@@ -347,18 +347,30 @@ library LoanLib {
...
+        uint256 interestPaid;
+
         if (toPay >= loan.interestOwed) {
-            toPay = toPay.sub(loan.interestOwed);
+            interestPaid = loan.interestOwed;
+            toPay = toPay.sub(interestPaid);
             loan.interestOwed = 0;
             if (toPay == 0) {
-                return;
+                return (0, interestPaid);
```
Juan-cc, line 364: Suggestion: the logic looks good to me so I'll approve it but I have a small suggestion. The logic is simple so it's not that complex to understand the meaning of this 0 value, but probably we could use a constant to describe the meaning of this 0 returned value if we want to be more descriptive (this also applies for other cases where we use 0 values).

Juan-cc (approved):
Approved with suggestion

passabilities (approved):

PR #254, opened 2021-01-04
[PROTOCOL-411][M01]: Pauser can prevent the caller from processing responses in a non-paused state

tpscrpt:
Started with the timelock part of the recommendation:
 - Add new setting name "TIMELOCK_SETTING"
 - Add new modifier (does nothing if no timelock, requires timelock passed otherwise)
 - Add new mapping to track timelocks (also in SettingsInterface)


salazarguille (approved):
Great job

PR #259, opened 2021-01-05
[PROTOCOL-424] N03 Gas optimization

sypeer:
It is:
- Removed unused or duplicated code


contracts/mock/util/AddressLibMock.sol
```diff
@@ -31,10 +31,6 @@ contract AddressLibMock {
         self.requireNotEmpty(REQUIRE_NOT_EMPTY_ERROR_MESSAGE);
     }
 
-    function requireEmpty(address self) external pure {
-        self.requireEmpty(REQUIRE_EMPTY_ERROR_MESSAGE);
-    }
-
```
passabilities, line 37: Need to readd as it is now being used by Chainlink

passabilities (changes requested):

contracts/mock/util/AddressLibMock.sol
sypeer: Changed ChainLinkAggregator to use a require with isEmpty() instead

passabilities (approved):

PR #262, opened 2021-01-07
[PROTOCOL-429] N08 - repeated or unused code

tpscrpt:
- The Base contract imports the BaseUpgradeable contract which uses the AddressLib library and the SettingsInterface interface. However, the Base contract reimports this library and interface unnecessarily.
- The SettingsInterface interface imported in the Base contract is never used.
- The isFinalized method from the NumberList library is never used in the entire code base.
- The ESCROW_FACTORY_LOGIC_NAME, CHAINLINK_AGGREGATOR_LOGIC_NAME, SETTINGS_LOGIC_NAME, LOGIC_VERSIONS_REGISTRY_LOGIC_NAME, MARKETS_STATE_LOGIC_NAME, ATM_FACTORY_LOGIC_NAME, and the ATM_SETTINGS_LOGIC_NAME are never used.

NOTE: I need some help sifting through the logic names to see which should be removed.

### DIFF
--- a/contracts/base/Base.sol
+++ b/contracts/base/Base.sol
@@ -3,13 +3,11 @@ pragma solidity 0.5.17;
 // Libraries
 import "@openzeppelin/contracts-ethereum-package/contracts/utils/ReentrancyGuard.sol";
 import "@openzeppelin/contracts-ethereum-package/contracts/utils/Address.sol";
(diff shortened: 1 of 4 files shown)


passabilities (approved):
The logic names aren't used for the most part in the code but I think are more of a reference point currently.

PR #264, opened 2021-01-08
[PROTOCOL-409] [C01]: Interest outcome can be manipulated

tpscrpt:
Fix adds a modifier to the Consensus contract requiring that the number of responses in the call is greater than or equal to number of singers - 1. Number of signers tracked when _addSigner and _removeSigner is called on OwnerSignersRole.

### DIFF
--- a/contracts/base/Consensus.sol
+++ b/contracts/base/Consensus.sol
@@ -60,18 +60,22 @@ contract Consensus is Base, OwnerSignersRole {
         _;
     }
 
+    /**
+        @notice Checks if the number of responses is greater or equal to a percentage of
+        the number of signers.
+     */
     modifier onlyEnoughSubmissions(uint256 responseCount) {
+        SettingsInterface settings = _getSettings();
+        uint256 percentageRequired = settings
+            .platformSettings(settings.consts().REQUIRED_SUBMISSIONS_PERCENTAGE_SETTING())
+            .value;
+        // https://stackoverflow.com/questions/17005364/dividing-two-integers-and-rounding-up-the-result-without-using-floating-point
+        uint256 minimumResponseCount = _signerCount.mul(percentageRequired).add(9999).div(
+            10000
+        );
+
         require(
-            responseCount >=
-                _signerCount
-                    .mul(
-                    _getSettings()
-                        .platformSettings(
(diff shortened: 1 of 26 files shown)


contracts/base/InterestConsensus.sol
```diff
@@ -44,14 +44,13 @@ contract InterestConsensus is InterestConsensusInterface, Consensus {
     function processRequest(
         TellerCommon.InterestRequest calldata request,
         TellerCommon.InterestResponse[] calldata responses
-    ) external isInitialized() isCaller(msg.sender) returns (uint256) {
-        require(
-            responses.length >=
-                _getSettings().getPlatformSettingValue(
-                    _getSettings().consts().REQUIRED_SUBMISSIONS_SETTING()
```
passabilities, line 51: Why are we not using the platform setting value anymore?

tpscrpt:
@passabilities Ready for review

contracts/base/Consensus.sol
```diff
@@ -60,6 +60,27 @@ contract Consensus is Base, OwnerSignersRole {
...
+     */
+    modifier onlyEnoughSubmissions(uint256 responseCount) {
+        SettingsInterface settings = _getSettings();
+        uint256 percentageRequired = settings
+            .platformSettings(settings.consts().REQUIRED_SUBMISSIONS_PERCENTAGE_SETTING())
+            .value;
+        // https://stackoverflow.com/questions/17005364/dividing-two-integers-and-rounding-up-the-result-without-using-floating-point
+        uint256 minimumResponseCount = _signerCount.mul(percentageRequired).add(9999).div(
+            10000
+        );
```
passabilities, line 75: I don't think this is correct:
```
signers = 6
percentRequired = 8000

signers * percentRequired + 9999 / 10000 = 48000.9999
```

We could use `NumbersLib.ratioOf()` to get the ratio of signers from the `responseCount` and then compare it to the platform setting. That way we don't have to worry about the rounding issue of getting 80% of 6 which would chop off the decimals.

Heres what I'm thinking:
```
uint256 signersRatio = responseCount.ratioOf(_signersCount);
// signersRatio = 6 * 10000 / 8 = 7500
require(signersRatio >= percentageRequired);
```

passabilities (changes requested):

passabilities (approved):

PR #274, opened 2021-01-12
[PROTOCOL-261]: How will lenders transfer the COMP tokens

tpscrpt:
Adds new function to LendingPool to:
 - Claim COMP tokens owed to the lending pool
 -  Swap COMP for LendingToken
 - Deposit swapped LendingToken into compound

Also includes new contracts:
 - ISwapper (interface for swappers like Uniswap)
 - ICompoundComptroller (compound comptroller interface for function we use -- claimComp)
 - UniswapController (to swap with using delegatecall, implementing ISwapper)
 - AssetController (to hold asset addresses, the uniswap controller and the compound comptroller)

And Settings.sol uses the new assetController contract.

Still untested.


contracts/base/LendingPool.sol
```diff
@@ -53,6 +60,8 @@ contract LendingPool is Base, LendingPoolInterface {
 
     MarketStateLib.MarketState internal compoundMarketState;
 
+    IERC20 private compToken;
```
passabilities, line 63: Lets not store it on the LendingPool but rather fetch it from the AssetsController via the mapping: `assetBySymbol('COMP')`

contracts/base/LendingPool.sol
```diff
@@ -423,4 +449,50 @@ contract LendingPool is Base, LendingPoolInterface {
         bool mintResult = tToken.mint(to, amount);
         require(mintResult, "TTOKEN_MINT_FAILED");
     }
+
+    /**
+        @notice It swaps the lending pool's accumualted COMP for the lendingToken
+        using Uniswap and deposits it into Compound.
+     */
+    function swapAccumulatedComp() external {
```
passabilities, line 457: Lets make this `internal`, return the value of **lendingTokens** claimed, and call it from both `_depositToCompound` and `_withdrawFromCompound`

1. When called from `_depositToCompound`, add the returned value of tokens claimed to the value of tokens that we wanted to deposit originally.
2. When called from `_withdrawFromCompound`, compare the returned valued of claimed tokens to the amount of tokens we are wanting to withdraw. 
    1. If we want to withdraw less than the amount claimed, call `_depositToCompound` for the remainder 
    2. If more than, proceed to withdraw the remaining value from compound
3. Check the balance of claimed tokens. Only swap if we actually claimed anything.
4. Can remove the `increaseSupply` call on market state since its being updated after the `_deposit` and `_withdraw` functions are called.

contracts/base/UniswapController.sol
```diff
@@ -0,0 +1,61 @@
...
+        external
+    {
+        require(
+            _minAmountOutPercent >= 8000 && _minAmountOutPercent < 10000,
+            "MIN_AMOUNT_INVALID"
+        );
+        require(
+            _maxSecondsToSwap > 0 && _maxSecondsToSwap <= 3600,
+            "MAX_SECONDS_INVALID"
+        );
```
passabilities, line 33: I'm pretty sure we don't need a `deadline` variable. Since we are calling the swap within the same transaction, we can just pass the `now` value and it will still work because the `now` value will never change in the same transaction.

contracts/base/UniswapController.sol
```diff
@@ -0,0 +1,61 @@
...
+        minAmountOutPercent = _minAmountOutPercent;
+        maxSecondsToSwap = _maxSecondsToSwap;
+        chainlink = _getSettings().chainlinkAggregator();
+    }
+
+    function swap(uint256 amountIn, address[] path) external returns (uint256 amountOut) {
+        IERC20 srcToken = IERC20(path[0]);
+        srcToken.safeIncreaseAllowance(ROUTER, amountIn);
+
+        uint256 chainlinkQuote = chainlink.valueFor(
```
passabilities, line 44: We will have to add support for the COMP token in our ChainlinkAggregator.

package.json
```diff
@@ -1,98 +1,91 @@
...
-    "ganache-core": "^2.11.2",
-    "truffle-hdwallet-provider": "^1.0.17",
-    "web3": "1.0.0-beta.55",
-    "web3-provider-engine": "15.0.0",
-    "web3-utils": "1.0.0-beta.55"
-  },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged"
+    "name": "teller-protocol-v1",
```
passabilities, line 2: what happened here? lets leave it as 2 spaces

passabilities (changes requested):

contracts/base/AssetController.sol
```diff
@@ -0,0 +1,51 @@
...
+    mapping(string => address) public assetBySymbol;
+
+    function getAssets() external view returns (address[] memory) {
+        return assets.array;
+    }
+
+    function addAsset(address asset) external {
+        assets.add(asset);
+        assetBySymbol[ERC20Detailed(asset).symbol()] = asset;
+    }
```
passabilities, line 28: Lets also update `Settings` to remove the `private` assets array and use this for adding/getting assets.

contracts/base/LendingPool.sol
tpscrpt: Forgot to delete that

package.json
tpscrpt: 2 spaces for all of JS? Sure

contracts/base/LendingPool.sol
tpscrpt: You're going to have to do 1. and 2. because I don't know how the other functions work or what you want to do with this.

contracts/base/UniswapController.sol
tpscrpt: No need for this now, using uniswap router's getAmounts() function

PR #275, opened 2021-01-14
Fixes Consensus Signers Count

passabilities:
It:
* Adds to the signers count in the right spot for consensus signers
* Removes the redundant IT initializer script


sypeer (approved):
LT

PR #277, opened 2021-01-17
[PROTOCOL-432] [PROTOCOL-438] [PROTOCOL-439] Update the LendingPool 

passabilities:
### Deploying TTokens In MarketFactory With LendingPool
The TTokens that are used for the protocol are now being created and deployed at the same time as the `LendingPool` in the `MarketFactory`. This enables us to dynamically create new pools by just specifying the token that the LendingPool will be using and not have to do anything extra off-chain.

### 1 LendingPool Per TToken
Before, we were creating multiple `LendingPool`s for each TToken meaning that the supply of the underlying token that was being lent out did not have a single source of truth as to the actual supply in the protocol. Now, in the `MarketFactory` it keeps a record of each `LendingPool` and uses the same one if another market is created that uses the same lending token but a different collateral token.

### MarketState Completely Remove
We were storing a record of the total supply of the lending token in a struct along with the total amount borrowed and total amount repaid for that asset. It is redundant to keep a record of the total value supplied since all we have to do is fetch the balance of the asset that the `LendingPool` has or, if it has a Compound token, fetch the underlying balance of the asset. By adding up these values all we have to do is keep a record of the total borrowed and total repaid values on the lending pool and we can calculate the "`MarketState`" in real time.


tpscrpt (approved):
Awesome PR! 🚀

sypeer (approved):

sypeer (approved):

PR #283, opened 2021-01-27
[PROTOCOL-440] Merge EtherCollateralLoans and TokenCollateralLoans into LoansBase

sypeer:
It is:
- Merging EtherCollateralLoans & TokenCollateralLoans into LoansBase


contracts/base/LoansBase.sol
```diff
@@ -622,4 +644,22 @@ contract LoansBase is LoansInterface, Base {
...
+    }
+
+    /**
+        @notice It transfers an amount of collateral tokens from an address to this contract.
+        @param from The address where the tokens will transfer from.
+        @param amount The amount to be transferred.
+     */
+    function _collateralTokenTransferFrom(address from, uint256 amount) internal {
+        ERC20Detailed(collateralToken).safeTransferFrom(from, address(this), amount);
+    }
```
passabilities, line 664: we can inline these functions

contracts/base/LoansBase.sol
```diff
@@ -582,6 +595,15 @@ contract LoansBase is LoansInterface, Base {
         @param amount The amount of collateral to be paid
      */
     function _payInCollateral(uint256 loanID, uint256 amount) internal {
+        if (collateralToken == _getSettings().ETH_ADDRESS()) {
+            // Ether collateral
+            require(msg.value == amount, "INCORRECT_ETH_AMOUNT");
```
passabilities, line 600: we should change the collateral token to be the `WETH` contract address that is on settings

also here is the point in which we should wrap the ETH value in WETH. after this point the only other point the contract needs to know if the collateral token is ETH or not is upon withdrawing. then we should be unwrapping the WETH and sending the caller back their unwrapped ETH collateral.

contracts/base/LoansBase.sol
```diff
@@ -554,7 +554,17 @@ contract LoansBase is LoansInterface, Base {
         uint256 loanID,
         uint256 amount,
         address payable recipient
-    ) internal;
+    ) internal {
+        totalCollateral = totalCollateral.sub(amount);
+        loans[loanID].collateral = loans[loanID].collateral.sub(amount);
+        if (collateralToken == _getSettings().ETH_ADDRESS()) {
+            // Ether collateral
+            recipient.transfer(amount);
```
passabilities, line 562: here we we unwrap the WETH and then transfer

passabilities (changes requested):

contracts/base/LoansBase.sol
sypeer: Done

contracts/base/LoansBase.sol
sypeer: Moved

contracts/base/LoansBase.sol
sypeer: done

PR #289, opened 2021-02-16
[PROTOCOL-466] Unit test ChainlinkAggregator contract functions

sypeer:
It is:
- Chainlink aggregator unit tests


test/chainlinkAggregator.test.ts
```diff
@@ -0,0 +1,128 @@
+import { Signer } from '@ethersproject/abstract-signer'
+import chai from 'chai'
+import chaiAsPromised from 'chai-as-promised'
+import hre from 'hardhat'
+import { ChainlinkAggregator } from '../types/typechain'
+import { getMarket } from '../tasks'
+import { Address } from '../types/custom/config-types'
+import { ETH_ADDRESS } from '../consts'
```
passabilities, line 8: for addresses we should stick to using the token addresses in the config instead of consts 
```typescript
import { getTokens } from '../config/tokens'
const tokens = getTokens(<Network>hre.network.name)
tokens.ETH
```

test/chainlinkAggregator.test.ts
```diff
@@ -0,0 +1,128 @@
...
+describe('Chainlink Aggregator', async () => {
+  let chainlinkAggregator: ChainlinkAggregator
+  let deployer: Signer
+  let sourceTokenAddress: Address
+  let destinationTokenAddress: Address
+  let aggregatorAddressToAdd: Address
+
+  // Setup for global tests
+  beforeEach(async () => {
+    deployer = await getNamedSigner('deployer')
```
passabilities, line 24: the deployer address won't change between tests

test/chainlinkAggregator.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  })
+
+  describe('Add an additional chainlink aggregator', () => {
```
passabilities, line 27: on the second nested `describe` lets start using the function name we are testing as the description
I.e. `describe('add', () => {})`

test/chainlinkAggregator.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    // Setup for snapshot tests
+    beforeEach(async () => {
+      // Get snapshot
+      await deployments.fixture('chainlink')
+      chainlinkAggregator = await contracts.get('ChainlinkAggregator', { from: deployer })
+
+      // Set up addresses
+      sourceTokenAddress = '0x7fc66500c84a76ad7e9c93437bfc5ac33e2ddae9' // AAVE
+      destinationTokenAddress = ETH_ADDRESS // ETH
+      aggregatorAddressToAdd = '0x6Df09E975c830ECae5bd4eD9d90f3A95a4f88012' // AAVE<>ETH
```
passabilities, line 37: these also won't change so they are unnecessarily being reset to the same values on every test

test/chainlinkAggregator.test.ts
```diff
@@ -0,0 +1,128 @@
...
```
passabilities, line 78: same

test/chainlinkAggregator.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+      // Add aggregator
+      await chainlinkAggregator.add(sourceTokenAddress, destinationTokenAddress, aggregatorAddressToAdd)
```
passabilities, line 81: for stuff like this where we need to revert back to a snapshot and then perform an action to set the environment up for the test, it would be a great use of the `createFixture` functionality that comes with the hardhat plugin. Essentially we can create our own custom fixtures within the test files.

https://hardhat.org/plugins/hardhat-deploy.html#creating-fixtures

test/chainlinkAggregator.test.ts
```diff
@@ -0,0 +1,128 @@
...
```
passabilities, line 118: same

passabilities (changes requested):

test/chainlinkAggregator.test.ts
sypeer: changed

test/chainlinkAggregator.test.ts
sypeer: done

test/chainlinkAggregator.test.ts
sypeer: renamed

test/chainlinkAggregator.test.ts
sypeer: got it, moved to initial setup

test/chainlinkAggregator.test.ts
sypeer: makes sense, added one here

test/chainlinkAggregator.test.ts
sypeer: done

test/chainlinkAggregator.test.ts
sypeer: done

passabilities (changes requested):
We will need additional unit tests for the `ChainlinkAggregator`

Also since these are unit tests, we should not be using the deployment snapshot that has already added aggregators. Only for integration tests should we be interacting with the fully deployed protocol

passabilities (approved):

PR #331, opened 2021-03-22
Gas Optimizations + LoanManager

passabilities:
A significant amount of code has been restructured in an effort to reduce the cost of transactions. 

### Dynamic Cache
The `DynamicProxy` contract has been given the ability to cache the implementation logic address so that it does not have to call the `LogicVersionsRegistry` contract for every external function call. By storing a cached logic address it enables several things:
1. Not having to check the `LogicVersionsRegistry` on every external function call for the implementation address
2. Have a block number buffer for caching the logic implementation
3. Have 2 types of `DynamicProxy` contracts: **strict** and **not strict**. A **strict** dynamic proxy means that it should always try to refresh its implementation address from the `LogicVersionsRegistry` (given the above caching rules). A **non strict** dynamic proxy meas that it should not try to refresh its implementation address but rather only allowed to upgrade its implementation address via a `upgradeProxyTo(address)` function that must be called by the logic version registry directly.

### LoanManager
The previous `Loans` contract has been renamed to `LoanManager`. The library `LoanLib` has been renamed to a contract `LoanData`. By doing so, if changes need to be made to the `LoanData` implementation, we do not need to deploy a new `LoanLib` and then a new `Loans`. Instead, just upgrade the `LoanData` contract. All of the functionality still works the same except the context of `LoanData` is now the same as `LoanManager` via a storage contract that holds all of the state variables for loans called `LoanStorage`. Calls to the `LoanData` contract are then called via a `delegatecall` so that it has access to the same loan values.

### Consensus 
The `LoanTermConsensus` contract was storing term submission data from a request in a mapping for every `createLoanWithTerms` call. This data was only ever being accessed within the same transaction. Because of this, we can not store the data in a mapping but rather create a local variable of the `struct` and pass it in memory. This has a **_significant_** gas savings of just under 50%

### TInitializeable 
For contracts that require to be deployed an then initialized, there was a modifier check to make sure that we cannot initialize the same contract more than once and another modifier that checks that the contract has been initialized. Because all of our contracts extend from `Base` that has a state variable for `Settings`, we can check that the contract has not already been initialized by checking that the settings address is null. Functions do not need to check that the contract has been initialized because for the most part, they would not even work if it was not. 

### Local Asset Settings
Some contracts need to access the asset settings contract and previously had to call the settings contract to then get the asset settings instance. To save gas costs, the asset setting instance has been saved in a state variable in `Base`. By making one less call, we can save gas.


tpscrpt (approved):
LGTM. Try to keep the PR sizes a bit smaller next time xP

sypeer (approved):

PR #332, opened 2021-03-23
Merge Consensus into LoanTermsConsensus

tpscrpt:


contracts/base/LoanTermsConsensus.sol
```diff
@@ -22,17 +27,75 @@ import "../interfaces/LoanTermsConsensusInterface.sol";
...
 
         It is used as rate limit per borrower address.
      */
-    mapping(address => uint256) public borrowerToLastLoanTermRequest;
+    mapping(address => uint256) public borrowerRequestTimes;
+
+    /**
+        @notice It tracks each request nonce value used by the borrower.
+     */
+    mapping(address => uint256) public borrowerNonces;
```
passabilities, line 51: we can get rid of these mappings and use the values from `LoanStorage` 
* `borrowerRequestTimes[borrower]` becomes `loans[borrowerLoans[borrower][borrowerLoans[borrower].length - 1].loanStartTime`
* `borrowerNonces[borrower]` becomes `borrowerLoans[borrower].length`

contracts/base/LoanTermsConsensus.sol
```diff
@@ -22,17 +27,75 @@ import "../interfaces/LoanTermsConsensusInterface.sol";
 
     @author develop@teller.finance
  */
-contract LoanTermsConsensus is LoanTermsConsensusInterface, Consensus {
-    /* Mappings */
+contract LoanTermsConsensus is LoanTermsConsensusInterface, OwnerSignersRole {
```
passabilities, line 30: - Bring logic and storage of signers over and can get rid of `_owner`. 
- Instead of using `Roles.Role` struct to keep track of the signers, we can use the `AddressArray` which means we can also get rid of `_signerCount` and just use `signers.array.length`

passabilities (changes requested):

passabilities (approved):

PR #343, opened 2021-05-01
Modify data types for more compact storage to reduce gas costs.

passabilities:
Before:
image

After:
image


contracts/storage/market.sol
```diff
@@ -12,86 +12,81 @@ import "../shared/libraries/NumbersList.sol";
...
 
 enum LoanStatus { NonExistent, TermsSet, Active, Closed, Liquidated }
 
 struct Loan {
-    // The id of the loan for internal tracking
-    uint256 id;
+    // Account that owns the loan
+    address payable borrower;
     // The asset lent out for the loan
     address lendingToken;
```
tpscrpt, line 29: Would it be possible to have loanId's be made of the borrower's address and the number of loans they have? I feel like that would remove 2 SLOADs in most cases.

tpscrpt (approved):
Clean changes, this was definitely needed.

contracts/storage/market.sol
passabilities: not if you want to do something with the loan and only know the ID, there would be no way to know who the loan belongs to

PR #365, opened 2021-05-26
Fork and deploy using new and existing deployments

passabilities:


deploy/nft.ts
```diff
@@ -18,7 +22,33 @@ const deployNFT: DeployFunction = async (hre) => {
     hre,
   })
 
-  const nftDistributor = await deployDiamond<ITellerNFTDistributor>({
+  let execute: DeployDiamondArgs<ITellerNFTDistributor, any>['execute']
+  try {
+    // Try to get deployment of TellerDiamond
+    await contracts.get('TellerNFTDistributor')
```
dumebi, line 28: This is also dependent on the TellerDiamond yes?

dumebi (approved):

sypeer (approved):

deploy/nft.ts
passabilities: yea, if it throws, then we deploy a new one, calling the `initialize` function as the `execute` parameter

PR #373, opened 2021-06-09
Features/story drivers

admazzola:
Architected the story testing system (roughed out)  

All classes are now defined with interaction and the flow of:

1) Generate stories (will be procedural) 
2) Loop through stories to generate Mocha Tests
3) Loop through and process the Mocha Tests (check them)

### DIFF
--- a/test/helpers/story/story-helpers.ts
+++ b/test/helpers/story/story-helpers.ts
@@ -12,29 +12,29 @@ import {
   LoanHelpersReturn,
   LoanDetailsReturn,
   CollateralFunctions,
-} from '../helpers/loans'
+} from '../loans'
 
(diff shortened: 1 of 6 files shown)


test/helpers/story/story-helpers.ts
```diff
@@ -12,29 +12,29 @@ import {
...
 
-export const LOAN_ACTIONS: string[] = [
-  'TAKE_OUT',
-  'LP_LEND',
-  'REPAY',
-  'LIQUIDATE',
-  'SWAP',
-  'LEND',
-]
+export const LOAN_ACTIONS = {
```
dumebi, line 30: Do we still need the loan actions? We already have story actions

dumebi (approved):

PR #390, opened 2021-06-16
NFT Interest Rate Platform Setting

passabilities:


contracts/settings/platform/libraries/PlatformSettingsLib.sol
```diff
@@ -27,6 +27,14 @@ library PlatformSettingsLib {
         return AppStorageLib.store().platformSettings[name];
```
dumebi, line 27: what does `s` mean in this context. wouldn't it be better to name the function `store` or something more descriptive?

dumebi (approved):

contracts/settings/platform/libraries/PlatformSettingsLib.sol
passabilities: just shorthand, but yes `store` is more descriptive. some lines of code started to get pretty long so we started doing this and just stuck with it.

mkhoury93 (approved):

PR #403, opened 2021-06-23
Feature/story test/master

dumebi:
We are building a systematic generator to test scenarios that could take place on the protocol. The story mode generator is built to randomly chain protocol actions and check out result
This generator is meant to be able to build out tests that have either been over looked or a way to apply some randomness to the tests.

our generated tests follow the format 
```
storyDomain  {
  domainName: 'DAPP.LEND.AAVE',
  scenarios: [
    { domain: 'DAPP', actions: [Array] },
    { domain: 'DAPP', actions: [Array] }
  ]
}
```

where each scenario is a permutated series of actions that could take place on a particular domain


admazzola:
This is great!  When I run 'yarn test mainnet' I do see 16 passing story tests and 6 failing.   However the file-based tests work well and I see the outputs of loans.tets.ts.  Perhaps those 6 can be skipped for now ? 

Some aspects that I don't fully understand are the 'TREE_STRUCTURE' constant so I think we need to talk about that and/or write documents on that.  I assume that will need to be appended to and it should to be as easy as possible to add a new 'domain' to this framework without having to hunt for multiple places to add new code.  

Secondly, I wonder if it would be possible to allow for manually specifying a story, so there will be some manually generated stories and some procedurally generated stories.   I think we should allow for this but perhaps the trickiest part will be integrating the 'args' in this way.  Maybe this can be an upgrade in the future or we can try and tackle it sooner rather than later.

dumebi:
> This is great!  When I run 'yarn test mainnet' I do see 16 passing story tests and 6 failing.   However the file-based tests work well and I see the outputs of loans.tets.ts.  Perhaps those 6 can be skipped for now ? 
> 
> 
> 
> Some aspects that I don't fully understand are the 'TREE_STRUCTURE' constant so I think we need to talk about that and/or write documents on that.  I assume that will need to be appended to and it should to be as easy as possible to add a new 'domain' to this framework without having to hunt for multiple places to add new code.  
> 
> 
> 
> Secondly, I wonder if it would be possible to allow for manually specifying a story, so there will be some manually generated stories and some procedurally generated stories.   I think we should allow for this but perhaps the trickiest part will be integrating the 'args' in this way.  Maybe this can be an upgrade in the future or we can try and tackle it sooner rather than later.   
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 

I agree with you. The tree structure aspect is the reworking of what we previously had to have one central point

I think a bit of code comment will suffice for it. And we also have our documentation to do on notion.

config/asset-settings.ts
```diff
@@ -9,7 +9,7 @@ const mainnetAssetSettings: AssetSettings = {
     { key: 'pPool', value: 'PCDAI', type: AssetType.Address },
     { key: 'MaxLoanAmount', value: 25000, type: AssetType.Amount },
     { key: 'MaxTVL', value: 10000000, type: AssetType.Amount },
-    { key: 'MaxDebtRatio', value: 5000, type: AssetType.Uint },
+    { key: 'MaxDebtRatio', value: 10000, type: AssetType.Uint },
```
sypeer, line 12: shouldn't we be testing with a more realistic debt ratio?

test/integration/escrow/dapps/compound.test.ts
```diff
@@ -42,51 +42,51 @@
...
 //           const { details } = await takeOut({
 //             lendToken: market.lendingToken,
 //             collToken: market.collateralTokens[0],
 //             loanType: LoanType.UNDER_COLLATERALIZED,
 //           })
-//
+
 //           const rando = await getNamedSigner('lender')
 //           await diamond
 //             .connect(rando)
```
sypeer, line 92: replace or delete please

config/asset-settings.ts
dumebi: this is from the previous settings we had i think. 

what's a more realistic `MaxDebtRatio`?

test/integration/escrow/dapps/compound.test.ts
dumebi: this particular test case?

config/asset-settings.ts
sypeer: I'd start with `5000`

test/integration/escrow/dapps/compound.test.ts
sypeer: yesir

PR #409, opened 2021-06-28
[PROTOCOL-634] [M01] Exchange rate rounding

dumebi:
added a check to ensure the calculated amounts to mint, redeem and burn are non-zero

### DIFF
--- a/contracts/lending/ttoken/TToken_V1.sol
+++ b/contracts/lending/ttoken/TToken_V1.sol
@@ -231,6 +231,8 @@ contract TToken_V1 is ITToken {
         // Calculate amount of tokens to mint
         uint256 mintAmount = _valueOfUnderlying(amount, exchangeRate());
 
+        require(mintAmount > 0, "Teller: amount to be minted cannot be 0");
+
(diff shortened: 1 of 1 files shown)


mkhoury93 (approved):

sypeer (dismissed):

contracts/lending/ttoken/TToken_V1.sol
```diff
@@ -284,6 +291,11 @@ contract TToken_V1 is ITToken {
         uint256 rate = exchangeRate();
         uint256 tokenValue = _valueOfUnderlying(amount, rate);
```
passabilities, line 292: In the interest of this comment

> Moreover, consider rounding up when calculating the [the number of tTokens to burn](https://github.com/teller-protocol/teller-protocol-v1/blob/cc80420ff73b0f7c466f6021ab57b2d71f96c3ba/contracts/lending/ttoken/TToken_V1.sol#L285).

lets append ` + 1`

passabilities (changes requested):

passabilities (approved):

PR #422, opened 2021-07-02
[PROTOCOL-657] Dynamically deploy contracts with network specific dapp addresses

sypeer:
The following facets have had constructor arguments added to save the network specific dapp addresses as an `immutable` variable:
- Sushiswap Facet 
- Uniswap Facet
- Aave Facet

LibUniswap and LibSushiswap have been merged into LibSwapper

TTokenAaveStrategy has an added init arg to store the LP address provider address and uses the updated LibDapps to grab the relevant Aave lending pool


config/index.ts
```diff
@@ -61,6 +62,8 @@ export const getNativeToken = (network: Network) => {
   return wrappedNativeToken
 }
 
+export const getDappAddresses = (network: Network) => dapps[network.name]
```
passabilities, line 65: should call `getNetworkName(network)`

deploy/markets.ts
```diff
@@ -129,6 +130,7 @@ const initializeMarkets: DeployFunction = async (hre) => {
                 case 'TokenSymbol':
                   return tokenAddresses.all[value]
                 case 'Address':
+                  return dappAddresses[value]
```
passabilities, line 133: this is good for now but will probably have to rethink things when we want to use addresses for other things

passabilities (changes requested):

config/index.ts
sypeer: fixed

passabilities (approved):

deploy/markets.ts
sypeer: can add additional types? `AddressProviderConstant` as an additional type

sypeer: Added an additional type `ProtocolAddressConstant` for now

dumebi (approved):

PR #443, opened 2021-07-15
unstake function

mkhoury93:
The Unstake function is an option for the user to unstake their NFTs on either Polygon or ETH MAINNET. This happens by removing them from the staked mapping then transferring to the users' respective wallets

### DIFF
--- a/contracts/nft/NFTFacet.sol
+++ b/contracts/nft/NFTFacet.sol
@@ -10,9 +10,7 @@ import { ADMIN, AUTHORIZED } from "../shared/roles.sol";
 import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 import { NFTLib } from "./libraries/NFTLib.sol";
 import { RolesLib } from "../contexts2/access-control/roles/RolesLib.sol";
-import {
-    EnumerableSet
-} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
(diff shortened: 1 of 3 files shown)


sypeer (commented):
nice! can we get a test to along with this?

mkhoury93:
> nice! can we get a test to along with this?

done

contracts/nft/NFTFacet.sol
```diff
@@ -44,6 +44,19 @@ contract NFTFacet is RolesMods {
...
 
+    /**
+     * @notice it unstakes the user's NFTs and transfers it from the diamond
+     * back to the user
+     * @param nftIDs the nftIDs to unstake
+     */
+    function unstakeNFTs(uint256[] calldata nftIDs) external {
+        for (uint256 i; i < nftIDs.length; i++) {
+            // Unstake NFTs
+            NFTLib.unstake(nftIDs[i]);
```
passabilities, line 55: we need checks to make sure these NFTs have been staked by the caller. you can do so by checking the boolean response value of the `unstake` library function like so:

```solidity
require(NFTLib.unstake(nftIDs[i]), "Teller: unstake NFT not owner");
```

also add these checks to the tests

passabilities (changes requested):

contracts/nft/NFTFacet.sol
mkhoury93: on it

PR #448, opened 2021-07-19
Fix NFT loan interest 

passabilities:
Loans with an NFT were given a default minimum interest ratio of 1%. The should only have a minimum of 1% if the interest rate defined on the loan was already > 0.


dumebi (approved):

mkhoury93 (dismissed):

sypeer (approved):

sypeer (approved):
Please make sure the failing dapp tests are fixed before merging in

PR #451, opened 2021-07-20
PROTOCOL-615: Feature/compound claim comp

dumebi:
Implemented `compoundClaimComp` function to claim accrued COMP


mkhoury93 (changes requested):
Maybe I am missing something, but i can't find the import of `CompoundClaimFacet` on the `ITellerDiamond`. We need to import it to the ITellerDiamond in order for the diamond to call functions on the facet

dumebi:
> Maybe I am missing something, but i can't find the import of `CompoundClaimFacet` on the `ITellerDiamond`. We need to import it to the ITellerDiamond in order for the diamond to call functions on the facet

fixed. surprisingly it worked without that

mkhoury93 (approved):

PR #456, opened 2021-07-28
Feature/manager profit fee

admazzola:
Adds an optional mgmt fee parameter to PlatformSettings, set at 5% by default.   When a loan is repaid, if excess profits were made with the borrowed escrowed lending token, then 5% of those excess profits will be given to the tToken liquidity pool such that 95% of the excess profits will be claimed back to the borrower. 

Unit tests were added to loans.tests.ts.


dumebi (dismissed):

test/integration/loans.test.ts
```diff
@@ -137,18 +162,420 @@ describe('Loans', () => {
...
+
+          const loanData = await diamond.getLoan(loanId)
+          console.log('loanData',  ) 
+
+         
+          console.log('balanceLeftToRepay 2',balanceLeftToRepay.toString())
+
+          console.log('balanceOf After', borrowerBalance.toString() )
+
+          expect(parseInt(borrowerBalance.toString())).to.equal(499980000)
```
passabilities, line 432: this works? looks like you are not updating this value after the loan is repaid. its value is only fetched before repayment on line 396

test/integration/loans.test.ts
```diff
@@ -137,18 +162,420 @@ describe('Loans', () => {
...
+          
+          const loanId =  helpers.details.loan.id 
+          console.log('loanId',loanId)
+
+        //  const lHelpers = await loanHelpers(loanId)  
+          
+
+
+          //need to give fake ERC20 to borrower 
+          //and add fake ERC20 to the borrow escrow
```
passabilities, line 469: is this to test profits with a different token? in order to calculate profits in the lending token, this token would have to have a chainlink aggregator registered via the ENS registry

test/integration/loans.test.ts
```diff
@@ -137,18 +162,420 @@ describe('Loans', () => {
...
+
+          const lendingTokenDecimals = await lendingToken.decimals()
+          console.log('decimals', lendingTokenDecimals)
+
+          await getFunds({
+            to: borrower,
+            tokenSym: market.lendingToken,
+            amount: 230 * 10**(lendingTokenDecimals),
+            hre,
+          })
```
passabilities, line 482: instead of getting funds for the borrower and then transferring the "fake profits" to the escrow below, you can just get the funds directly to the loan escrow address for the lending token. this should make it easier to calculate the proper balance that the borrower should have after the profit fee is taken

test/integration/loans.test.ts
```diff
@@ -137,18 +162,420 @@ describe('Loans', () => {
...
+         
+          console.log('balanceLeftToRepay 2',balanceLeftToRepay.toString())
+
+          console.log('balanceOf After', borrowerBalance.toString() )
+
+
+
+          tTokenEscrowBalance = await lendingToken.balanceOf(tTokenAddress)
+           console.log('tTokenEscrowBalance 2 ', tTokenEscrowBalance.toString())
+           expect(tTokenEscrowBalance).to.equal(99801540000)  //99700030000
```
passabilities, line 562: instead of hardcoding this value, because this value depends on how much was originally in the TToken, lets grab the TToken balance before and after the repay. The difference should be `the loan repayment + profit fee`

passabilities (changes requested):

contracts/settings/platform/libraries/PlatformSettingsLib.sol
```diff
@@ -95,6 +95,14 @@ library PlatformSettingsLib {
         value_ = s(NAMES.LIQUIDATE_REWARD_PERCENT).value;
     }
 
+    /**
+     * @notice It gets the current "ManagementFeePercent" setting's value
+     * @return value_ the current value.
+     */
+    function getManagementFeePercent() internal view returns (uint256 value_) {
+        value_ = s(NAMES.MANAGEMENT_FEE_PERCENT).value;
```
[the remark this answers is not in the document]
admazzola: okay I did that.  Warning: all of the other ones still return uint256 so platformsettingslib.sol is inconsistent looking now

passabilities:
@admazzola this PR branch can be deleted as well correct?

ethereumdegen:
yes - now we are to use feature/manager-profit-fee-r2.  I did that because starting fresh was easier than rebasing this branch

PR #480, opened 2021-08-19
[PROTOCOL-954] Take Out Loan V3

mkhoury93:
Given that the user has a combination of V1 and V2 tokens, we need to combine both NFT ID types to take out a loan. It doesn't make sense for a user to discriminate between a V1 and a V2 token when they are just trying to take out a loan.

An extra version # has been added in the `CreateLoanWithNFTFacet.sol` and the `MainnetCreateLoanWithNFTFacet.sol`. A new test has also been added in the `loans.test.ts` and the helper function for `takeOutLoanWithNFTs` is updated with the new `case: 3`

### DIFF
--- a/contracts/market/CreateLoanWithNFTFacet.sol
+++ b/contracts/market/CreateLoanWithNFTFacet.sol
@@ -125,11 +125,12 @@ contract CreateLoanWithNFTFacet is ReentryMods, PausableMods {
         uint16 version,
         bytes memory tokenData
     ) internal virtual returns (uint256 allowedLoanSize_) {
-        if (version == 2) {
-            (uint256[] memory nftIDs, uint256[] memory amounts) = abi.decode(
-                tokenData,
-                (uint256[], uint256[])
-            );
+        if ((2 & version) == 2) {
+            (
+                uint256[] memory nftIDsV1,
+                uint256[] memory nftIDs,
+                uint256[] memory amounts
+            ) = abi.decode(tokenData, (uint256[], uint256[], uint256[]));
             for (uint256 i; i < nftIDs.length; i++) {
                 NFTLib.applyToLoanV2(loanID, nftIDs[i], amounts[i], msg.sender);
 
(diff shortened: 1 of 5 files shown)


contracts/market/CreateLoanWithNFTFacet.sol
```diff
@@ -115,6 +115,23 @@ contract CreateLoanWithNFTFacet is ReentryMods, PausableMods {
                     version,
                     tokenData_
                 );
+        } else if (version == 3) {
+            // get loan size from first function
+            uint256 loanSize1 = CreateLoanWithNFTFacet._takeOutLoanProcessNFTs(
```
dumebi, line 120: if this is for allowed loan size, isn't it better to name them `allowedLoanSize1`, `allowedLoanSize2`?

dumebi (changes requested):

contracts/market/CreateLoanWithNFTFacet.sol
```diff
@@ -115,6 +115,23 @@ contract CreateLoanWithNFTFacet is ReentryMods, PausableMods {
                     version,
                     tokenData_
                 );
+        } else if (version == 3) {
```
passabilities, line 118: lets leave this function to just have this (remove the `if` and `else`)
```solidity
            return _takeOutLoanProcessNFTs(loanID, version, tokenData_);
```

contracts/market/CreateLoanWithNFTFacet.sol
```diff
@@ -146,6 +163,26 @@ contract CreateLoanWithNFTFacet is ReentryMods, PausableMods {
                 );
                 allowedLoanSize_ += baseLoanSize * amounts[i];
             }
+        } else if (version == 3) {
```
passabilities, line 166: remove this `else if` from here. we will add the additional checks we need in the `MainnetCreateLoanWithNFTFacet`

contracts/market/mainnet/MainnetCreateLoanWithNFTFacet.sol
```diff
@@ -34,6 +34,20 @@ contract MainnetCreateLoanWithNFTFacet is CreateLoanWithNFTFacet {
                     nftIDs[i]
                 );
             }
+        } else if (version == 3) {
```
passabilities, line 37: change this to the bitwise operations we went over, where the version is 3 we:

```solidity
uint256 v1Amount = ...;
uint256 v2Amount = super._takeOutLoanProcessNFTs(...);
allowedLoanSize_ = v1Amount + v2Amount;
```

passabilities (changes requested):

passabilities (approved):

dumebi (approved):

PR #495, opened 2021-10-15
added poly and mainnet diamond extensions

elclandestin0:
It creates an extension for both mainnet and polygon networks

### DIFF
--- a/deploy/protocol.ts
+++ b/deploy/protocol.ts
@@ -225,38 +225,15 @@ const deployProtocol: DeployFunction = async (hre) => {
     execute,
   }
 
-  const mainnetDiamondArgs: DeployDiamondArgs<IMainnetDiamond, any> = {
-    hre,
-    name: 'MainnetDiamond',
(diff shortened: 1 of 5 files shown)


deploy/protocol.ts
```diff
@@ -222,15 +223,27 @@ const deployProtocol: DeployFunction = async (hre) => {
...
+    hre,
+    name: 'MainnetDiamond',
+    facets,
+    owner: deployer,
+    execute,
+  }
   const diamond = await deployDiamond<ITellerDiamond, any>(tellerDiamondArgs)
+  const mainnetDiamond = await deployDiamond<IMainnetDiamond, any>(
+    mainnetDiamondArgs
+  )
```
passabilities: What about Polygon,

passabilities (changes requested):

dumebi (changes requested):
can you check the failed tests, seems a file is missing

PR #499, opened 2021-10-27
[PROTOCOL-1202] Chainlinkn use latestRoundData instead of latestAnswer

sypeer:
It is:
- Replacing latestAnswer with latestRoundData for the freshest data
- Added extra validations

### DIFF
--- a/contracts/price-aggregator/pricers/ChainlinkPricer.sol
+++ b/contracts/price-aggregator/pricers/ChainlinkPricer.sol
@@ -40,7 +40,7 @@ contract ChainlinkPricer {
         ) = ChainlinkAgg(getEthAggregator(token)).latestRoundData();
         require(rawPrice > 0, "Chainlink price <= 0");
         require(updateTime != 0, "Incomplete round");
-        require(answeredInRound >= roundID, "Stale price");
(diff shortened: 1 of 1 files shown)


contracts/price-aggregator/pricers/ChainlinkPricer.sol
```diff
@@ -31,9 +31,17 @@ contract ChainlinkPricer {
...
+        (
+            uint80 roundID,
+            int256 rawPrice,
+            uint256 startedAt,
+            uint256 updateTime,
+            uint80 answeredInRound
+        ) = ChainlinkAgg(getEthAggregator(token)).latestRoundData();
+        require(rawPrice > 0, "Chainlink price <= 0");
+        require(updateTime != 0, "Incomplete round");
+        require(answeredInRound >= roundID, "Stale price");
```
passabilities: Lets check the docs from Chainlink. I want to give these checks a 1 or 2 round buffer in case so the app doesn't break if a round isn't completed in time but want to make sure that is an acceptable thing to do

passabilities (dismissed):

dumebi (approved):

passabilities (approved):

PR #500, opened 2021-10-27
[PROTOCOL-1267] Enabling eth deposits

sypeer:


passabilities (changes requested):
We'll also need to deposit into WETH as we are doing when depositing ETH as collateral in the `LibCollateral.deposit` function

contracts/lending/LendingFacet.sol
```diff
@@ -70,15 +74,23 @@ contract LendingFacet is RolesMods, ReentryMods, PausableMods {
...
+                amount
+            );
+            // Set allowance for Teller token to pull funds to mint
+            SafeERC20.safeIncreaseAllowance(
+                IERC20(asset),
+                address(tToken),
+                amount
+            );
+        } else {
+            ERC20WrapperUpgradeable(asset).depositFor(address(this), amount);
```
passabilities: Lets create an `immutable` constant during facet deployment for the WETH address and `require(asset == WETH)` to ensure that we are correctly depositing ETH to the correct contract.

Also you need to modify this line to use [this interface](https://github.com/teller-protocol/teller-protocol-v1/blob/develop/contracts/shared/interfaces/IWETH.sol#L19):
```solidity
IWETH(asset).depositFor{ value: msg.value }();
```

passabilities (changes requested):

contracts/lending/LendingFacet.sol
sypeer: why not use the ERC20Wrapper interface?

passabilities: **ERC20WrapperUpgradeable**
image

**IWETH**
image

sypeer: using the `wrappedNativeToken` address from AppStorage instead of adding another immutable constant https://github.com/teller-protocol/teller-protocol-v1/blob/develop/contracts/storage/app.sol#L34

contracts/lending/LendingFacet.sol
```diff
@@ -70,15 +72,27 @@ contract LendingFacet is RolesMods, ReentryMods, PausableMods {
...
+                msg.sender,
+                address(this),
+                amount
+            );
+            // Set allowance for Teller token to pull funds to mint
+            SafeERC20.safeIncreaseAllowance(
+                IERC20(asset),
+                address(tToken),
+                amount
+            );
```
passabilities: Lastly, since the end result of checking it there is a `msg.value` and depositing into WETH is that we for sure have an ERC20 token now to deal with, this block of code should be after the `if/else` and before the `safeTransfer` below as the `tToken.mint` function needs to pull in the tokens deposited

passabilities (changes requested):

passabilities (approved):

PR #521, opened 2021-12-22
Boni

sypeer:


contracts/lending/ttoken/strategies/compound/TTokenCompoundStrategy_2.sol
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * @dev it creates a reference to the Compound storage
+     */
+    function() pure returns (CompoundStorage.Store storage)
+        private constant compoundStore = CompoundStorage.store;
+
+    string public constant NAME = "CompoundStrategy_2";
+
+    uint32 constant BLOCKS_PER_YEAR = 2351330;
```
passabilities: since we are working with **timestamps** lets use the solidity untis of `1 years` which will give us the number of seconds there are per year. you can remove this **constant** variable

contracts/lending/ttoken/strategies/compound/TTokenCompoundStrategy_2.sol
```diff
@@ -0,0 +1,205 @@
...
+
+    function totalUnderlyingSupply() external override returns (uint256) {
+        // Get current supply
+        uint256 currentSupply = tokenStore().underlying.balanceOf(
+            address(this)
+        ) + compoundStore().cToken.balanceOfUnderlying(address(this));
+        // Calculate bonus interest due
+        uint256 bonusInterest = ((currentSupply * 10) / 100) *
+            ((block.timestamp - compoundStore().lastBonusIntTimestamp) /
+                BLOCKS_PER_YEAR);
```
passabilities: here is where you should replace with `1 years`

contracts/lending/ttoken/strategies/compound/TTokenCompoundStrategy_2.sol
```diff
@@ -0,0 +1,205 @@
...
+        uint256 bonusSupply = tokenStore().underlying.balanceOf(
+            compoundStore().bonusGnosisSafe
+        );
```
passabilities: can remove (not, and shouldn't be, used)

- actually only case to use the supply of the gnosis safe is if `bonusInterest > bonusSupply` in which case we should use `bonusInterest = bonusSupply`

contracts/lending/ttoken/strategies/compound/TTokenCompoundStrategy_2.sol
```diff
@@ -0,0 +1,205 @@
...
+        // Deposit into
+        tokenStore().underlying.transferFrom(
+            compoundStore().bonusGnosisSafe,
+            address(this),
+            bonusInterest
+        );
+        ITToken_V3(address(this)).mint(bonusInterest);
```
passabilities: we actually need to remove this line. by not minting any additional TTokens and depositing additional underlying assets, we inherently are generating additional APY for existing lenders

contracts/lending/ttoken/strategies/compound/TTokenCompoundStrategy_2.sol
```diff
@@ -0,0 +1,205 @@
...
+
+        return
+            tokenStore().underlying.balanceOf(address(this)) +
+            compoundStore().cToken.balanceOfUnderlying(address(this));
```
passabilities: instead of duplicating this code (first 2 lines of the function) we can do this instead:
```solidity
return currentSupply + bonusInterest;
```

contracts/lending/ttoken/strategies/compound/TTokenCompoundStrategy_2.sol
```diff
@@ -0,0 +1,205 @@
...
+import { LibMeta } from "../../../../shared/libraries/LibMeta.sol";
+import {
+    LibCompound
+} from "../../../../escrow/dapps/libraries/LibCompound.sol";
+
+// Storage
+import "../../token-storage.sol" as TokenStorage;
+import "./compound-storage.sol" as CompoundStorage;
+
+contract TTokenCompoundStrategy_2 is RolesMods, TTokenStrategy {
```
passabilities: In order to showcase the fact that we are only modifying the behavior of `totalUnderlyingSupply`, lets extend from `TTokenCompoundStrategy` and only modify the `totalUnderlyingSupply` function and `constructor`.

By doing this, in the `totalUnderlyingSupply` function you can simply do this:
```solidity
        uint256 currentSupply = super.totalUnderlyingSupply();
```
instead of:
```solidity
        uint256 currentSupply = tokenStore().underlying.balanceOf(
            address(this)
        ) + compoundStore().cToken.balanceOfUnderlying(address(this));
```

passabilities (changes requested):

contracts/lending/ttoken/strategies/compound/TTokenCompoundStrategy_2.sol
sypeer: `years` has been depreciated as a uint denomination. using `365 days`

contracts/lending/ttoken/strategies/compound/TTokenCompoundStrategy_2.sol
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * @notice Sets the address of the bonus gnosis safe for the bonus interest disbursment
+     * @param _bonusGnosisSafe The address of the gnosisSafe to pull funds from for bonus interest
+     */
+    constructor(address _bonusGnosisSafe) {
+        bonusGnosisSafe = _bonusGnosisSafe;
+        compoundStore().lastBonusIntTimestamp = block.timestamp;
+        NAME = "CompoundStrategy_2";
```
passabilities: state will not be carried over as this contract is `delegatecall`ed by the TToken contract, therefore, these 2 values we set in the constructor won't be recognized by the TToken storage. 

We will need to move these value initialization into the `initialize` function again unfortunately

deploy/markets.ts
```diff
@@ -102,11 +102,22 @@ const initializeMarkets: DeployFunction = async (hre) => {
...
-      indent: 4,
-    })
+    let tTokenStrategy: Contract
+
+    if (market.strategy.name == 'TTokenCompoundStrategy_2') {
+      tTokenStrategy = await deploy({
+        hre,
+        contract: market.strategy.name,
+        indent: 4,
+        args: [await (await getNamedSigner('deployer')).getAddress()],
```
passabilities: we will also need to set an approval amount on the underlying asset to allow the TToken contract to pull funds from the address we set here

test/helpers/set-test-env.ts
```diff
@@ -158,6 +162,26 @@ const fundMarket = async (testEnv: TestEnv) => {
...
+    // Approve protocol for bonus int
+    await token
+      .connect(deployer)
+      .approve(tToken.address, bnedAmount.mul(10))
+      .then(({ wait }) => wait())
+
+    const allowance = await token.allowance(
+      await deployer.getAddress(),
+      tToken.address
+    )
```
passabilities: this needs to be moved to the `market` deploy script so that the allowance is set on deployment, not just the tests.

Will need to add a network check to prevent calling `getFunds` on any non-`hardhat` networks.

Also, only set allowance if the current `allowance === 0`

test/lending/bonusIntStrategy.ts
```diff
@@ -0,0 +1,68 @@
...
+
+chai.should()
+chai.use(solidity)
+
+setTestEnv('Lending - TToken bonus interest', (testEnv: TestEnv) => {
+  it.only('Should accurately dispense the bonus interest', async () => {
+    // Load tToken
+    const { tokens, tellerDiamond, lender, hre } = testEnv
+
+    const dai: ERC20 = tokens.find((o) => o.name === 'DAI')!.token
```
passabilities, line 19: to accurately test that this works, we need to test with a fresh deployment

passabilities (changes requested):

contracts/lending/ttoken/strategies/compound/TTokenCompoundStrategy_2.sol
```diff
@@ -0,0 +1,68 @@
...
+        uint256 dailyInterest = ((currentSupply / 10) / 365 days) * 1 days;
+        uint256 bonusInterest = dailyInterest * (interestTimeperiod / 1 days);
+
+        uint256 gnosisBalance = tokenStore().underlying.balanceOf(
+            bonusGnosisSafe
+        );
+        if (gnosisBalance < bonusInterest) {
+            bonusInterest = gnosisBalance;
+        }
+        if (bonusInterest > 0) {
```
passabilities: I think we should move line 26-31 inside the `if` on line 32
This will prevent unnecessary calls to underlying `balanceOf` function if we don't need to

deploy/markets.ts
```diff
@@ -102,15 +103,35 @@ const initializeMarkets: DeployFunction = async (hre) => {
...
     })
+    // Setting approval for bonus int from deployer account
+    if (market.strategy.name == 'TTokenCompoundStrategy_2') {
+      const lendingToken = await contracts.get<IERC20>('IERC20', {
+        at: lendingTokenAddress,
+      })
+      await lendingToken
+        .connect(await getNamedSigner('deployer'))
+        .approve(tTokenAddress, toBN('1000000', '18'))
+    }
```
passabilities: we should only set the approval if its currently set to 0

deploy/markets.ts
```diff
@@ -102,15 +103,35 @@ const initializeMarkets: DeployFunction = async (hre) => {
...
-      indent: 4,
-    })
+    let tTokenStrategy: Contract
+
+    if (market.strategy.name == 'TTokenCompoundStrategy_2') {
+      tTokenStrategy = await deploy({
+        hre,
+        contract: market.strategy.name,
+        indent: 4,
+        args: [await (await getNamedSigner('deployer')).getAddress()],
```
passabilities: we should add a named account to the `hardhat.config.ts`

test/helpers/set-test-env.ts
```diff
@@ -158,6 +162,15 @@ const fundMarket = async (testEnv: TestEnv) => {
       amount: bnedAmount,
       hre,
     })
+
+    // Get funds for deployer for bonus int
+    await getFunds({
+      to: deployer,
```
passabilities: use the named account here

passabilities (changes requested):

deploy/markets.ts
sypeer: added as `gnosisSafe`

passabilities (approved):

PR #528, opened 2025-10-23
Claim NFT task print args

passabilities:
* Adds `--print-args` flag to `claim-nft` task to use on Etherscan
  * E.g. `hh claim-nft --network mainnet --account 0x123 --print-args`
* Optimization to `view-nft` task to view un/claimed nfts


ethereumdegen:
The new code looks good to me. I was able to set up my local environment and run the task as well.

